PR #19, opened 2021-07-08
Enable TLS

ishustava:
Fixes hashicorp/consul-ecs#16

This PR enables TLS for the consul on ECS components:
- Enable TLS for the consul dev server, keeping HTTP ports still open
- Enable TLS for the consul client, also keeping HTTP port still open. 
- ~~Use consul-ecs image that enables envoy to talk over TLS with the consul client~~ (we no longer use TLS between envoy and consul client)
- Refactor mesh-task module and create iam roles there, allowing users to provide additional policies to attach to that role
- Add an acceptance test testing TLS enabled.


modules/mesh-task/main.tf
```diff
@@ -54,14 +54,136 @@ locals {
...
+    {
+      "Effect": "Allow",
+      "Action": [
+        "ecs:ListTasks",
+        "ecs:DescribeTasks"
+      ],
+      "Resource": [
+        "*"
+      ]
+    },
```
ishustava, line 88: I believe this is is only needed if dev server is enabled for the `discover-servers` command.

test/acceptance/tests/basic/basic_test.go
```diff
@@ -29,120 +30,131 @@ func TestValidation_RetryJoinRequired(t *testing.T) {
...
-				"--cluster",
-				suite.Config().ECSClusterARN,
-				"--family",
-				fmt.Sprintf("consul_server_%s", suite.Config().Suffix),
-			},
-		})
-		r.Check(err)
+	for _, secure := range cases {
+		t.Run(fmt.Sprintf("secure: %t", secure), func(t *testing.T) {
+			randomSuffix := strings.ToLower(random.UniqueId())
```
ishustava, line 37: Changing it to creating a random suffix for each test because otherwise we can't re-run tests locally for the same ECS cluster from `setup-terraform`. That is because AWS doesn't delete secrets permanently; it just schedules them for deletion. So if you re-run any test that has tls enabled, your second run will fail because AWS will think you're trying to create a secret that's scheduled for deletion.

modules/dev-server/main.tf
```diff
@@ -173,7 +243,51 @@ exec consul agent -server \
   -hcl 'telemetry { disable_compat_1.9 = true }' \
   -hcl 'connect { enabled = true }' \
   -hcl 'enable_central_service_config = true' \
+%{if var.tls~}
+  -hcl='ca_file = "/consul/consul-agent-ca.pem"' \
+  -hcl='cert_file = "/consul/dc1-server-consul-0.pem"' \
+  -hcl='key_file = "/consul/dc1-server-consul-0-key.pem"' \
+  -hcl='auto_encrypt = {allow_tls = true}' \
+  -hcl='ports { https = 8501 }' \
```
lkysow, line 251: should we set `http = -1`?

modules/dev-server/outputs.tf
```diff
@@ -12,3 +12,11 @@ output "lb_security_group_id" {
   description = "Security group ID of load balancer in front of Consul server."
   value       = var.lb_enabled ? aws_security_group.load_balancer[0].id : null
 }
+
+output "ca_cert_arn" {
```
lkysow, line 16: need descriptions

modules/dev-server/main.tf
ishustava: I don't see a reason to do that since it's a dev server. It makes our changes much easier in that we don't need to change the load balancer. Since we're only using it for testing, we thought it's not worth the complexity.

modules/dev-server/variables.tf
```diff
@@ -73,3 +73,9 @@ variable "assign_public_ip" {
   type        = bool
   default     = false
 }
+
+variable "tls" {
+  description = "Whether to enable TLS on the server."
```
lkysow, line 78: Could be expanded to explain it's for Consul control plane comms to differentiate from Connect.

modules/mesh-task/main.tf
lkysow: correct

modules/mesh-task/main.tf
```diff
@@ -54,14 +54,136 @@ locals {
...
+resource "aws_iam_role_policy_attachment" "additional_task_policies" {
+  count      = length(var.additional_task_role_policies)
+  role       = aws_iam_role.task.id
+  policy_arn = var.additional_task_role_policies[count.index]
+}
+
+resource "aws_iam_policy" "execution" {
+  name        = "${var.family}-execution"
+  path        = "/ecs/"
+  description = "Consul mesh-task execution policy"
```
lkysow, line 117: Should this include `var.family`?

modules/mesh-task/templates/consul_client_command.tpl
```diff
@@ -17,3 +21,10 @@ exec consul agent \
   -hcl 'ports { grpc = 8502 }' \
   -hcl 'advertise_reconnect_timeout = "15m"' \
   -hcl 'enable_central_service_config = true' \
+%{ if tls ~}
+  -hcl 'ca_file = "/tmp/consul-ca-cert.pem"' \
+  -hcl 'auto_encrypt = {tls = true}' \
+  -hcl "auto_encrypt = {ip_san = [\"$ECS_IPV4\"]}" \
+  -hcl 'verify_outgoing = true' \
+  -hcl 'ports {https = 8501}' \
```
lkysow, line 29: disable http?

modules/mesh-task/variables.tf
```diff
@@ -3,14 +3,16 @@ variable "family" {
   type        = string
 }
 
-variable "execution_role_arn" {
-  description = "ARN for the task's execution role. This role is used internally by ECS."
-  type        = string
+variable "additional_task_role_policies" {
+  description = "List of additional policy ARNs to attach to the task role."
+  type        = list(any)
```
lkysow, line 8: `list(string)`?

modules/mesh-task/variables.tf
```diff
@@ -34,7 +36,7 @@ variable "consul_image" {
 variable "consul_ecs_image" {
   description = "consul-ecs Docker image."
   type        = string
-  default     = "docker.mirror.hashicorp.services/hashicorp/consul-ecs:0.1.2"
+  default     = "ishustava/consul-ecs-dev:latest"
```
lkysow, line 39: heads up that I've found docker hub images heavily rate limited and ended up using ghcr.

modules/mesh-task/variables.tf
```diff
@@ -87,3 +89,15 @@ variable "tags" {
   type        = map(string)
   default     = {}
 }
+
+variable "tls" {
+  description = "Whether to enable TLS for the mesh-task."
```
lkysow, line 94: Maybe explain specifically this is for control plane traffic. Nit: not sure we have to refer to "the mesh-task" since this is a param for it so it feels slightly redundant.

modules/mesh-task/variables.tf
```diff
@@ -87,3 +89,15 @@ variable "tags" {
...
+  type        = bool
+  default     = false
+}
+
+variable "consul_server_ca_cert_arn" {
+  description = "The ARN of the secret of the Consul server CA certificate."
```
lkysow, line 100: Should we add something to `validation.tf` that this is set if `tls` is true?

lkysow (approved):
Awesome work.

modules/dev-server/main.tf
lkysow: Oh yeah totally agreed!

pglass (approved):

test/acceptance/tests/basic/basic_test.go
```diff
@@ -29,120 +30,131 @@ func TestValidation_RetryJoinRequired(t *testing.T) {
...
+				consulServerTaskARN = tasks.TaskARNs[0]
+			})
 
-	var tasks listTasksResponse
-	require.NoError(t, json.Unmarshal([]byte(taskListOut), &tasks))
-	require.Len(t, tasks.TaskARNs, 1)
+			// Wait for both tasks to be registered in Consul.
+			retry.RunWith(&retry.Timer{Timeout: 3 * time.Minute, Wait: 10 * time.Second}, t, func(r *retry.R) {
+				out, err := shell.RunCommandAndGetOutputE(t, shell.Command{
+					Command: "aws",
```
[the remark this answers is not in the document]
lkysow: I think when I wrote this it probably didn't support `exec` cuz it was a new feature. Is this supported now?

modules/mesh-task/templates/consul_client_command.tpl
ishustava: Resolving since we've decided to only have HTTP.

test/acceptance/tests/basic/basic_test.go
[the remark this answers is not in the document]
pglass: Ah, I see. Yeah, the Go SDK has [ExecuteCommand](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ecs#Client.ExecuteCommand) now, but it's not nice to use... It returns an [ssm.Session](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ecs@v1.6.0/types#Session) with a `StreamURL` (websocket) which seems to [use a custom binary format](https://stackoverflow.com/a/64248831). And I couldn't find an easy example of how to read from it in Golang (and didn't dig any further).

modules/mesh-task/templates/consul_client_command.tpl
```diff
@@ -3,10 +3,17 @@ ECS_IPV4=$(curl -s $ECS_CONTAINER_METADATA_URI | jq -r '.Networks[0].IPv4Address
...
+echo "$CONSUL_CACERT" > /tmp/consul-ca-cert.pem
+%{ endif }
+
 exec consul agent \
   -advertise "$ECS_IPV4" \
   -data-dir /consul/data \
   -client 0.0.0.0 \
+  -hcl 'addresses = { dns = "127.0.0.1" }' \
+  -hcl 'addresses = { grpc = "127.0.0.1" }' \
+  -hcl 'addresses = { http = "127.0.0.1" }' \
```
ishustava, line 16: @lkysow @pglass adding these for the Consul client so that we only bind to localhost for those APIs since we don't have TLS.

test/acceptance/tests/basic/basic_test.go
```diff
@@ -22,127 +23,150 @@ func TestValidation_RetryJoinRequired(t *testing.T) {
 		TerraformDir: "./terraform/retry-join-validate",
 		NoColor:      true,
 	})
-	defer terraform.Destroy(t, terraformOptions)
+	defer terraform.DestroyE(t, terraformOptions)
```
ishustava, line 26: Needed to add this because it was failing for me locally because `terraform.Destroy` calls `require.NoError` and `terraform destroy` would have the same error as the apply. I'm not sure why it works in CI (perhaps a different version of TF?)

modules/mesh-task/main.tf
```diff
@@ -115,11 +237,6 @@ resource "aws_ecs_task_definition" "this" {
                 hostPort      = 8300
                 protocol      = "udp"
               },
-              {
-                containerPort = 8500
-                hostPort      = 8500
-                protocol      = "tcp"
-              },
```
ishustava, line 122: Removing since nobody should need to talk to the consul client outside the mesh task and so it should expose its http port on the host.

modules/mesh-task/variables.tf
ishustava: Added!

modules/dev-server/main.tf
```diff
@@ -42,36 +93,44 @@ resource "aws_ecs_task_definition" "this" {
...
+            containerPath = "/consul"
+          }
+        ]
+        linuxParameters = {
+          initProcessEnabled = true
         }
-      ]
-      linuxParameters = {
-        initProcessEnabled = true
+        dependsOn = var.tls ? [
```
lkysow, line 126: should we update containerPorts on line 111

modules/dev-server/main.tf
```diff
@@ -173,7 +243,51 @@ exec consul agent -server \
...
+  -hcl='verify_server_hostname = true' \
+%{endif~}
 EOF
+
+  consul_server_tls_init_command = <<EOF
+ECS_IPV4=$(curl -s $ECS_CONTAINER_METADATA_URI | jq -r '.Networks[0].IPv4Addresses[0]')
+cd /consul
+echo "$CONSUL_CACERT" > ./consul-agent-ca.pem
+echo "$CONSUL_CAKEY" > ./consul-agent-ca-key.pem
+consul tls cert create -server -additional-ipaddress=$ECS_IPV4
```
lkysow, line 263: Worth a comment saying we need this init container to generate the server cert because the IP of the task must be included in the SAN and we don't know the IP ahead of time.

modules/mesh-task/main.tf
```diff
@@ -54,14 +54,136 @@ locals {
...
+        "ssmmessages:CreateDataChannel",
+        "ssmmessages:OpenControlChannel",
+        "ssmmessages:OpenDataChannel"
+      ],
+      "Resource": [
+        "*"
+      ]
+    }
+  ]
+}
```
lkysow, line 103: At some point we'll make this optional since it's only needed for `aws exec`

modules/mesh-task/validation.tf
```diff
@@ -1,4 +1,5 @@
 locals {
   require_retry_join               = (var.consul_server_service_name == "" && var.retry_join == "") ? file("ERROR: either consul_server_service_name or retry_join must be set so that Consul clients can join the cluster") : null
   require_port_unless_no_port_true = !var.outbound_only && var.port == 0 ? file("ERROR: port must be set if outbound_only is false") : null
+  require_ca_cert_if_tls_enabled   = (var.tls && var.consul_server_ca_cert_arn == "") ? file("ERROR: Consul CA certificate is required when TLS is enabled") : null
```
lkysow, line 4: Error should reference the name of the input variable
```suggestion
  require_ca_cert_if_tls_enabled   = (var.tls && var.consul_server_ca_cert_arn == "") ? file("ERROR: consul_server_ca_cert_arn must be set if tls is true") : null
```

lkysow (approved):
Great work!

modules/dev-server/main.tf
ishustava: I need to keep that one because the load balancer needs it in either tls or no-tls case. We could probably make it conditional on whether lb is enabled, but since it's the dev server I'm not sure if it matters. wdyt?

modules/mesh-task/main.tf
ishustava: Should we add a GH issue for that?

modules/dev-server/main.tf
lkysow: yeah leave it but I was thinking we'd add port 8501 to it

modules/mesh-task/main.tf
lkysow: I added it to Asana

modules/dev-server/main.tf
ishustava: We don't really need it though, nothing is connecting to it over that host port.

pglass (approved):

modules/dev-server/main.tf
ishustava: I'm going to resolve this but let me know if you think we need that host IP, and I can add it in a separate PR.

lkysow: Forgot that the clients hit it over RPC, not the HTTPS endpoint. Resolved!

PR #20, opened 2021-07-16
Use output from setup-terraform by default to create config to run acceptance tests

ishustava:
Since we use `setup-terraform` template to set up our acceptance tests, I thought it'd be nice if tests just took the output of that terraform and parsed them into the test config. This makes running tests locally much easier. Let me know what you think!

This PR also adds a few other minor changes:
- Skip cleanup if both `-no-cleanup-on-failure` is set and if test has failed.
- Removes `-suffix` flag. We were no longer using it after the change to generate random ids for each test in #19

### DIFF
--- a/test/acceptance/framework/flags/flags.go
+++ b/test/acceptance/framework/flags/flags.go
@@ -1,7 +1,10 @@
 package flags
 
 import (
+	"encoding/json"
 	"flag"
+	"fmt"
+	"os/exec"
 	"sync"
 
 	"github.com/hashicorp/terraform-aws-consul-ecs/test/acceptance/framework/config"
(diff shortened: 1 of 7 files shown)


test/acceptance/framework/flags/flags.go
```diff
@@ -44,25 +49,65 @@ func (t *TestFlags) init() {
...
+		if err != nil {
+			return nil, err
+		}
+		err = json.Unmarshal(cmdOutput, &tfOutput)
+		if err != nil {
+			return nil, err
+		}
+
+		for k, v := range tfOutput {
+			if k == "private_subnets" {
```
lkysow, line 85: comment about why subnets need to be handled specially would be nice

test/acceptance/tests/basic/terraform/ca-cert-validate/main.tf
```diff
@@ -0,0 +1,14 @@
...
+module "test_client" {
+  source = "../../../../../../modules/mesh-task"
+  family = "family"
+  container_definitions = [{
+    name = "basic"
+  }]
+  outbound_only              = true
+  consul_server_service_name = "test"
+  tls                        = true
+}
```
lkysow, line 14: does this pass t fmt?

lkysow (commented):
I think it looks good but might need a rebase to ensure I'm reviewing just the changes pertaining to this.

ishustava:
> I think it looks good but might need a rebase

Sorry! Let me rebase!

test/acceptance/framework/flags/flags.go
ishustava: I've added more comments to this code and also switched the output to be just 'subnets' in `setup-terraform`. I think it makes it easier for tests and since that TF is only used by tests I think it's OK.

test/acceptance/tests/basic/terraform/ca-cert-validate/main.tf
ishustava: yep, it does.

lkysow (approved):
🎉

pglass (approved):
Nice!

PR #24, opened 2021-08-18
Use AWS CloudMap to discover servers via DNS

ishustava:


iftachk:
Very nice!

modules/dev-server/main.tf
```diff
@@ -247,6 +251,25 @@ resource "aws_iam_role" "this_task" {
   }
 }
 
+resource "aws_service_discovery_private_dns_namespace" "server" {
+  name        = "consul"
+  description = "The namespace for the Consul dev server."
```
pglass, line 256: nit: Include `var.name` in the description to help distinguish namespaces by cluster

pglass (approved):
Looks great! Good work, Iryna!

lkysow (approved):

modules/dev-server/main.tf
ishustava: I looked at prefixing it with name, but then the dns name looks kind of repetitive. So for example, if `var.name` is `foo`, then the DNS name will look like `foo.foo-consul`. For our example terraform, it'll look like `foo-consul-server.foo-consul-server-consul`. I think I'll instead make it a var in the dev-server module and default to `consul` so that people can set it to whatever they want.

PR #25, opened 2021-08-23
Consul ECS on EC2 example

pglass:


lkysow (approved):
It looks good. Did you want us to test it too?

pglass:
> It looks good. Did you want us to test it too?

It would be nice, but not required for an approval, I guess, since it's not scoped for beta anyway. I've tested it several times myself already.

PR #30, opened 2021-08-30
fix: added support for adding custom tags that append to existing value

### LINKED ISSUE #29: Add ability to specify tags to task definitions
In the current Terraform configuration the tag value of `consul.hashicorp.com/mesh` is hard coded. In order to append to this value the Terraform template must be changed to accept tags provided through the `tags` variable. 

```hcl
  tags = {
    "consul.hashicorp.com/mesh" = "true"
  }
```

karl-cardenas-coding:
This PR adds support for custom tags while retaining the default value of `"consul.hashicorp.com/mesh" = "true"`. In addition, the output for displaying all provided tags is also added. 

Closes #29

### DIFF
--- a/modules/mesh-task/outputs.tf
+++ b/modules/mesh-task/outputs.tf
@@ -1,3 +1,8 @@
 output "task_definition_arn" {
   value = aws_ecs_task_definition.this.arn
 }
+
+output "task_tags" {
+  value = aws_ecs_task_definition.this.tags_all
+}
+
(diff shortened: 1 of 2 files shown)


pglass (approved):

modules/mesh-task/outputs.tf
```diff
@@ -1,3 +1,8 @@
 output "task_definition_arn" {
   value = aws_ecs_task_definition.this.arn
 }
+
+output "task_tags" {
```
lkysow, line 5: do we need this output? Wondering how you would consume it.

karl-cardenas-coding: It dosen't hurt to expose it. It could be used in terratest cases to validate that the resource has the required tags.

PR #31, opened 2021-08-31
Add ACL controller module and enable ACLs for other components.

ishustava:
The new `controller` module runs ACL controller implemented in consul-ecs
for managing ACL for ECS tasks on the service mesh.

This PR also enables ACLs for the `dev-server` and the `mesh-task` modules.

Note for the reviewers:
There are a couple of future improvements that will come in a later PR:
1. Handling error cases when setting different ACL related vars (e.g. service name prefix should be required if acls are true)
2. Possibly making polling interval(s) for the controller configurable
3. Allowing users to bring their own tokens for services in case they don't want to run the controller.


test/acceptance/tests/basic/basic_test.go
```diff
@@ -128,6 +128,31 @@ func TestBasic(t *testing.T) {
 			require.NoError(t, json.Unmarshal([]byte(taskListOut), &tasks))
 			require.Len(t, tasks.TaskARNs, 1)
 
+			// Create an intention.
```
pglass, line 131: Q: This intention is needed because...we specify a default deny when ACLs are enabled?

ishustava: Yep, exactly. When ACL default policy is "default-deny" intentions are also "default-deny" and so we need to explicitly add one. 

It's also mentioned in the [Intentions](https://www.consul.io/docs/connect/intentions#intentions) docs but it's a bit non-obvious:

> The default intention behavior is defined by the default ACL policy. If the default ACL policy is "allow all", then all Connect connections are allowed by default. If the default ACL policy is "deny all", then all Connect connections or requests are denied by default.

modules/acl-controller/README.md
```diff
@@ -0,0 +1,6 @@
+# ACL Controller
```
lkysow, line 1: Do we forsee any other uses for this? Wondering if we should just call it controller.

modules/acl-controller/main.tf
```diff
@@ -0,0 +1,178 @@
+resource "aws_secretsmanager_secret" "client_token" {
+  name = "${var.secret_name_prefix}-consul-client-token"
+}
+
+resource "aws_secretsmanager_secret_version" "client_token" {
+  secret_id     = aws_secretsmanager_secret.client_token.id
+  secret_string = jsonencode({})
+}
+
+resource "aws_ecs_service" "consul-controller" {
```
lkysow, line 10: resource names are underscore right?

modules/acl-controller/main.tf
```diff
@@ -0,0 +1,178 @@
...
+  name            = "consul-acl-controller"
```
lkysow, line 11: Should we support the name prefix?

lkysow: actually I just realized I've spent the last 2 years trying to remove prefixes from k8s, nvm 😂

modules/acl-controller/variables.tf
```diff
@@ -0,0 +1,47 @@
+variable "consul_ecs_image" {
+  description = "consul-ecs Docker image."
+  type        = string
+  default     = "ishustava/consul-ecs-dev:acl-test"
```
lkysow, line 4: reminder to change before merge

modules/mesh-task/variables.tf
```diff
@@ -100,3 +100,21 @@ variable "gossip_key_secret_arn" {
   type        = string
   default     = ""
 }
+
+variable "acls" {
```
lkysow, line 104: should this be `acls_enabled`?

lkysow (commented):
LGTM. Just a question on naming.

test/acceptance/tests/basic/basic_test.go
ishustava: Actually, I lied. I discovered that default-deny needs to be set on both clients and servers for it to be the case, and I was only setting it on servers. I'll update to make clients use default-deny as well

ishustava: I'll update acceptance test as well to check that we can't connect unless we add an intention.

modules/acl-controller/README.md
ishustava: We can potentially re-use the code in consul-ecs but I feel like the tf module is specific to ACLs in terms of the variables it takes and outputs it produces. It's also hard to predict what we'll do/need in the future, but I think if we need to rename it and/or make it more generic, it should definitely possible.

For now, I'm inclined to keep it specific. WDYT?

modules/acl-controller/main.tf
ishustava: I think we should, but only for some resources. Currently none of them are prefixed, and I ran into some problems with that the other day. There are two types or resources for which we'd need a prefix:
1. IAM policies and roles are global. If we don't prefix them we could potentially overwrite a role of another controller running in a different cluster. So those need to be prefixed.
2. Task definitions are shared as well, meaning they don't belong to any specific cluster. So we'd need to prefix the task definition. 

Q: Should we also prefix the service for consistency? I think it'd be kind of nice if it showed just "consul-acl-controller" in the UI without the prefix. The downside is that when you look at tasks, you'll see that the task def is prefixed, and so there will be a disconnect there perhaps.

modules/mesh-task/variables.tf
ishustava: Right now this is mirroring `tls` var for consistency. I like that they are consistent but don't have strong preference for which one is better. Should I change both to `{tls,acls}_enabled`?

modules/acl-controller/main.tf
pglass: > Should we also prefix the service for consistency?

Either way. I lean slightly toward yes.

Since we are adding a name prefix, can we use it for the secret name prefix too and remove the `secret_name_prefix` variable? The name prefix should be unique enough, since it's used to name globally unique IAM resources.

pglass (approved):

lkysow (approved):
Sorry, I missed this.

PR #33, opened 2021-09-10
Acceptance tests for EC2

pglass:
- [x] Permissions fix so CircleCI jobs can cleanup instances: https://github.com/hashicorp/consul-test-environments/pull/90
- [x] Update to run EC2/Fargate in parallel (waiting to push this until CI cleanup is working)


test/acceptance/setup-terraform/ec2.tf
```diff
@@ -0,0 +1,64 @@
...
+
+  tags = merge(var.tags, {
+    Name = local.name
+  })
+}
+
+resource "aws_ecs_cluster" "ec2" {
+  name = "${local.name}-ec2"
+  tags = var.tags
+}
```
pglass, line 64: In order to run Fargate and EC2 tests in parallel, I create a separate cluster for the EC2 tests. 

Technically, a single cluster with both Fargate and EC2 capacity could work. But ECS clusters are free and it felt safer with separate clusters.

test/acceptance/tests/basic/basic_test.go
```diff
@@ -30,134 +30,178 @@ func TestValidation_CACertRequiredIfTLSIsEnabled(t *testing.T) {
...
+		TerraformDir: "./terraform/basic-install",
+		Vars:         tfVars,
+		NoColor:      true,
+		EnvVars:      map[string]string{},
+	})
+	terraform.Init(t, terraformOptions)
+
+	// Unique-ify state files by launch type since we parallelize by launch type.
+	// This means only two state files to deal with in case of manual cleanup locally.
+	stateFile := fmt.Sprintf("terraform-%s.tfstate", launchType)
```
pglass, line 88: Calling this out: separate Terraform state files in order to run terraform in parallel.

terratest has a [CopyTerraformFolderToTemp](https://pkg.go.dev/github.com/gruntwork-io/terratest/modules/test-structure#CopyTerraformFolderToTemp) function, that can copy the `basic-install` module to a temp dir. But, it doesn't fix relative module paths, and makes manual cleanup difficult ("where's that temp dir with the Terraform state file?")

test/acceptance/setup-terraform/ec2.tf
pglass: Oh, and `setup-terraform` will always setup both EC2 instances as-is. Do we want a variable so we can choose between Fargate or EC2 or both?

test/acceptance/framework/config/config.go
```diff
@@ -2,23 +2,23 @@ package config
...
-	ECSClusterARN      string `json:"ecs_cluster_arn"`
-	Subnets            interface{}
-	Suffix             string
-	Region             string
-	VpcID              string `json:"vpc_id"`
-	LogGroupName       string `json:"log_group_name"`
-	Tags               interface{}
+	NoCleanupOnFailure    bool
+	ECSClusterARN_Fargate string `json:"ecs_cluster_arn_fargate"`
+	ECSClusterARN_EC2     string `json:"ecs_cluster_arn_ec2"`
```
ishustava, line 7: what if we rename them a bit to remove underscores and also keep readable, e.g `FargateClusterARN` and `EC2ClusterARN`?

test/acceptance/setup-terraform/ec2.tf
ishustava: Initially, I was imagining that we would have two CI jobs, and each of them would run either only Fargate or EC2. That way we can have one blueprint that we can run in multiple configurations. 

I think the main question here to help decide is if we will have other tests that we'd want to test on both of these launch types (e.g any of the gateways). In that case, I think it'd be easier to configure a CI job and pass some flag to tests to run them in either configuration rather than adding the test for each launch type for each feature. WDYT?

ishustava (commented):
Just took a brief look, but wanted to leave a question about the approach before I review more in detail.

test/acceptance/setup-terraform/ec2.tf
pglass: Yeah, I went for test parallelization at the golang level so that we can run tests locally in parallel, too. I can't run tests locally with just a CircleCI yaml (or actually, [maybe we can?](https://circleci.com/docs/2.0/local-cli/))

In any case, we can split these into separate acceptance test jobs in CircleCI:
* The CircleCI jobs would run `setup-terraform` and pass in the launch type (which would be a new variable).
* The CircleCI jobs would select EC2/Fargate with test filters: `go test -run 'TestEC2.*' ...`.

>  In that case, I think it'd be easier to configure a CI job and pass some flag to tests to run them in either configuration rather than adding the test for each launch type for each feature.

🤔  I'm tired right now, so maybe not thinking this through - but adding golang test functions which I can run locally seems easier than needing to configure/test a new CI job for each new feature? We can definitely discuss post-Beta release!

pglass: @ishustava Okay, I've thought about this some more and I agree! It should be simpler, and I don't think I'd actually want to run tests for all launch types locally all the time anyway.

test/acceptance/framework/flags/flags.go
```diff
@@ -48,6 +50,7 @@ func (t *TestFlags) init() {
 		"If true, the tests will not clean up resources they create when they finish running."+
 			"Note this flag must be run with -failfast flag, otherwise subsequent tests will fail.")
 	flag.StringVar(&t.flagECSClusterARN, flagECSClusterARN, "", "ECS Cluster ARN.")
+	flag.StringVar(&t.flagLaunchType, flagLaunchType, "", "The ECS launch type to test: Fargate or EC2.")
```
lkysow, line 53:
```suggestion
	flag.StringVar(&t.flagLaunchType, flagLaunchType, "", "The ECS launch type to test: 'FARGATE' or 'EC2'.")
```

Also should this default to fargate?

test/acceptance/setup-terraform/variables.tf
```diff
@@ -13,3 +13,20 @@ variable "tags" {
...
+}
+
+variable "instance_count" {
+  description = "Number of EC2 instances to create for the EC2 launch type (if enabled)."
+  type        = number
+  default     = 3
+}
+
+variable "instance_type" {
+  description = "The instance type for EC2 instances."
```
lkysow, line 29:
```suggestion
  description = "The instance type for EC2 instances if launch type is EC2."
```

lkysow (approved):

test/acceptance/setup-terraform/ec2.tf
```diff
@@ -0,0 +1,58 @@
...
+  count = local.ec2_enabled ? 1 : 0
+
+  name = local.name
+  role = aws_iam_role.instance_role[count.index].name
+}
+
+resource "aws_instance" "instances" {
+  count = local.ec2_enabled ? var.instance_count : 0
+
+  ami                  = local.esc_optimized_ami
```
pglass, line 43: fix typo `esc` -> `ecs`

PR #40, opened 2021-09-23
Add acceptance tests to check that ACL token is deleted

ishustava:


test/acceptance/tests/basic/terraform/basic-install/main.tf
```diff
@@ -44,7 +44,7 @@ variable "secure" {
 variable "consul_ecs_image" {
   description = "Consul ECS image to use."
   type        = string
-  default     = "docker.mirror.hashicorp.services/hashicorpdev/consul-ecs:latest"
+  default     = "ishustava/consul-ecs-dev:delete-acl"
```
ishustava, line 47: Will revert once hashicorp/consul-ecs#30 is merged

test/acceptance/tests/basic/basic_test.go
```diff
@@ -82,6 +83,49 @@ func TestBasic(t *testing.T) {
 				consulServerTaskARN = tasks.TaskARNs[0]
 			})
 
+			t.Cleanup(func() {
```
lkysow, line 86: It feels a bit weird to run this test inside of `t.Cleanup`. Maybe there's prior art here but I thought we usually use Cleanup to clean up things and it's relatively small.

What if we had a separate test for this functionality?

ishustava: Let me see if I can move it out cleanup. 

There is a cost to adding a new test just for this because it'll add quite a bit of execution time as the setup is the same. Ideally, these assertions are just part of the end-to-end service mesh workflow.

lkysow (approved):

test/acceptance/tests/basic/basic_test.go
```diff
@@ -132,77 +116,53 @@ func TestBasic(t *testing.T) {
...
 				r.Check(err)
 				if !strings.Contains(curlOut, `"code": 200`) {
 					r.Errorf("response was unexpected: %q", curlOut)
 				}
 			})
 
+			// Check that the ACL tokens for services are deleted
+			// when services are destroyed.
+			if secure {
+				if suite.Config().NoCleanupOnFailure && t.Failed() {
```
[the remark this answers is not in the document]
ishustava: Ah good point! I'll remove it.

test/acceptance/tests/basic/basic_test.go
```diff
@@ -132,77 +116,53 @@ func TestBasic(t *testing.T) {
...
+					terraform.Destroy(t, terraformOptions)
+
+					// Check that the ACL tokens are deleted from Consul.
+					retry.RunWith(&retry.Timer{Timeout: 5 * time.Minute, Wait: 20 * time.Second}, t, func(r *retry.R) {
+						out, err := executeRemoteCommand(t, consulServerTaskARN, "consul-server", `/bin/sh -c "consul acl token list"`)
+						require.NoError(r, err)
+						require.NotContains(r, out, fmt.Sprintf("test_client_%s", randomSuffix))
+						require.NotContains(r, out, fmt.Sprintf("test_server_%s", randomSuffix))
+					})
+					terraformOptions.Targets = nil
```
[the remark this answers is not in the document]
ishustava: I've updated as you suggested. I agree, I think it's better create a new struct.

test/acceptance/tests/basic/basic_test.go
```diff
@@ -211,3 +171,26 @@ func TestBasic(t *testing.T) {
 type listTasksResponse struct {
 	TaskARNs []string `json:"taskArns"`
 }
+
+// executeRemoteCommand executes a command inside a container in the task specified
+// by taskARN.
```
pglass, line 176: 🥳 

Future improvement: we can have this trim the excess output from the session-manager-plugin:

```
The Session Manager plugin was installed successfully. Use the AWS CLI to start a session.

Starting session with SessionId: ecs-execute-command-0b27d055bac783a77
...

Exiting session with sessionId: ecs-execute-command-0b27d055bac783a77.
```

Then we could parse JSON output from commands easily (instead of resorting to string search or regex).

pglass (approved):
Looks great! 🎉

test/acceptance/tests/basic/terraform/basic-install/main.tf
ishustava, line 47:
```suggestion
  default     = "docker.mirror.hashicorp.services/hashicorpdev/consul-ecs:latest"
```

PR #44, opened 2021-09-27
Add pull request template

pglass:


.github/pull_request_template.md
```diff
@@ -0,0 +1,15 @@
+Changes proposed in this PR:
```
erichaberkorn, line 1: It might be clearer once this is filled out to make each of the headings bold

.github/pull_request_template.md
```diff
@@ -0,0 +1,15 @@
...
+
+How I've tested this PR:
+
+How I expect reviewers to test this PR:
+
+Checklist:
+- [ ] Tests added
+- [ ] CHANGELOG entry added 
+
+  * HashiCorp engineers only, community PRs should not add a changelog entry.
```
erichaberkorn, line 13: Are wanting these lines in the final PR description. If they are meant to be removed, I think we can do something like this.
```suggestion
[//]: # (  * HashiCorp engineers only, community PRs should not add a changelog entry. )
```

erichaberkorn (approved):
This looks good to me. It feels like a wonder starting place for us to iterate on :shipit:

.github/pull_request_template.md
pglass: Nice! This seemed to work for a multiline comment in the markdown:

```
[//]: # ( 
   HashiCorp engineers only. Community PRs should not add a changelog entry.
   Changelog entries should use present tense, e.g. "Add support for..."
)
```

lkysow: The comment is supposed to be so that community contribs don't try and add a changelog entry. So it needs to somehow be associated to that checkbox imo

pglass: > it needs to somehow be associated to that checkbox imo

Experimented a little (based on [this SO question](https://stackoverflow.com/a/20885980)) and the following seems to work.

```
- [ ] CHANGELOG entry added 

```

You do need a blank line prior to the `[comment]::` lines. It also affects the vertical spacing of the bullet points...but it does exclude those comments from rendering.

lkysow (approved):

ishustava (approved):

PR #62, opened 2021-11-05
Allow passing volumes into mesh-task

pglass:
- Add a `volumes` variable to mesh-task so that users can define additional volumes for the ECS task

## How I've tested this PR:
Unit test included

👀 

## Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/modules/mesh-task/main.tf
+++ b/modules/mesh-task/main.tf
@@ -207,6 +207,70 @@ resource "aws_ecs_task_definition" "this" {
     name = local.consul_binary_volume_name
   }
 
+  dynamic "volume" {
+    for_each = var.volumes
+    content {
+      name      = volume.value["name"]
+      host_path = lookup(volume.value, "host_path", null)
+
+      dynamic "docker_volume_configuration" {
+        for_each = contains(keys(volume.value), "docker_volume_configuration") ? [
+          volume.value["docker_volume_configuration"]
+        ] : []
+        content {
+          autoprovision = lookup(docker_volume_configuration.value, "autoprovision", null)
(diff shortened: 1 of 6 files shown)


modules/mesh-task/main.tf
```diff
@@ -207,6 +207,70 @@ resource "aws_ecs_task_definition" "this" {
     name = local.consul_binary_volume_name
   }
 
+  dynamic "volume" {
```
pglass, line 210: If there is a better way to do this, let me know.

It's translating a list of maps (`volumes = [{...}]`) into several blocks (`volume {}`). From what I can tell, Terraform doesn't support passing blocks into modules. And, it doesn't support dynamic names for the keys in these dynamic blocks.

erichaberkorn: Eek, this doesn't look fun. Did this come from somewhere else? If so, it would be great to link to the source.

erichaberkorn (approved):

modules/mesh-task/main.tf
pglass: > Did this come from somewhere else?

No, I crafted this myself. I followed the [Terraform docs on dynamic blocks](https://www.terraform.io/docs/language/expressions/dynamic-blocks.html) and referred to the [volumes field in the `aws_ecs_task_definition`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecs_task_definition#volume) and looked at a some examples of nested dynamic blocks.

(One way we maybe might avoid this is to implement mesh-task as a Terraform provider, but that's essentially a rewrite that we'd need to think through.)

ishustava (approved):
Sorry for the delay!

PR #63, opened 2021-11-10
Fix a secrets manager deletion and recreation bug

erichaberkorn:
Since the ACL-controller/mesh task rely on deterministic secret names, we need the ability to delete a secret and then recreate it. Without this change deleting and recreating a secrete got this error: "Error: error creating Secrets Manager Secret: InvalidRequestException: You can't create this secret because a secret with this name is already scheduled for deletion."

## How I've tested this PR:
With the ACLs enabled:
1. Create a mesh task service
2. Delete a mesh task service
3. Recreate the mesh task service

👁️ 

## Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,11 @@ IMPROVEMENTS
   shut down gracefully. [[GH-48](https://github.com/hashicorp/terraform-aws-consul-ecs/pull/48)]
   [[GH-61](https://github.com/hashicorp/terraform-aws-consul-ecs/pull/61)]
 
+BUG FIXES
+* modules/acl-controller and modules/mesh-task: Fix a bug that results in
+  AWS Secrets Manager secrets failing to be created.
+  [[GH-63](https://github.com/hashicorp/terraform-aws-consul-ecs/pull/63)]
+
 ## 0.2.0-beta2 (Sep 30, 2021)
 
 FEATURES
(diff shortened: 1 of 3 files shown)


lkysow (approved):

pglass:
Q: Is this still the test you ran?

> Ensure tasks are running for the service
> Stop all tasks for that service
> Ensure tasks are running for the service

If so, this won't get the ACL controller to actually delete the AWS Secret. The controller only "clears" the secret, [by updating it with an empty json object](https://github.com/hashicorp/consul-ecs/blob/main/controller/resource.go#L281-L284) during a delete. 

The AWS Secrets themselves are created by [mesh-task](https://github.com/hashicorp/terraform-aws-consul-ecs/blob/main/modules/mesh-task/main.tf#L183-L192). So test, we could do a `terraform apply; terraform destroy; terraform apply` (with the same resource names). (Unless I'm misunderstanding!)

erichaberkorn:
> Q: Is this still the test you ran?
> 
> > Ensure tasks are running for the service
> > Stop all tasks for that service
> > Ensure tasks are running for the service
> 
> If so, this won't get the ACL controller to actually delete the AWS Secret. The controller only "clears" the secret, [by updating it with an empty json object](https://github.com/hashicorp/consul-ecs/blob/main/controller/resource.go#L281-L284) during a delete.
> 
> The AWS Secrets themselves are created by [mesh-task](https://github.com/hashicorp/terraform-aws-consul-ecs/blob/main/modules/mesh-task/main.tf#L183-L192). So test, we could do a `terraform apply; terraform destroy; terraform apply` (with the same resource names). (Unless I'm misunderstanding!)

@pglass Sorry about that. You are correct. I'll update the description.

PR #67, opened 2021-11-12
Optional application entrypoint override

pglass:
Allow users an easy way to override the application entrypoint for better graceful shutdown.

* Add a variable `application_shutdown_delay_seconds`. When provided, a custom entrypoint is used for the application containers, which will delay the TERM signal for a number of seconds. 
* The option is ignored for any container definition which has the `entryPoint` field set.

## How I've tested this PR:
Acceptance tests

👀 🧠 

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 


test/acceptance/tests/basic/terraform/basic-install/main.tf
```diff
@@ -159,20 +159,7 @@ module "test_client" {
...
-function onterm() {
-    echo "TEST LOG: Caught sigterm. Sleeping 10s..."
-    sleep 10
-    exit 0
-}
-trap onterm TERM
-wait $PID
-EOT
-      ]
+      command = ["/app/fake-service"]
```
pglass, line 162: This part is not so seamless. Since this overrides the entrypoint, users will have to provide a `command` to specify how their container starts up - unless there's a default command in the image).

I haven't found a way at runtime to discover the default entrypoint from the image.

test/acceptance/tests/basic/terraform/basic-install/main.tf
```diff
@@ -49,7 +49,7 @@ variable "launch_type" {
 variable "consul_ecs_image" {
   description = "Consul ECS image to use."
   type        = string
-  default     = "docker.mirror.hashicorp.services/hashicorpdev/consul-ecs:latest"
+  default     = "docker.mirror.hashicorp.services/hashicorpdev/consul-ecs:49ccc3a"
```
pglass, line 52:
```suggestion
  default     = "docker.mirror.hashicorp.services/hashicorpdev/consul-ecs:latest"
```

To fix before merging

test/acceptance/tests/basic/terraform/basic-install/main.tf
lkysow: Should document that requirement in the variable docs?

modules/mesh-task/variables.tf
```diff
@@ -170,6 +170,12 @@ variable "acl_secret_name_prefix" {
 
 variable "consul_datacenter" {
   type        = string
-  description = "The name of the Consul datacenter the client belongs to"
+  description = "The name of the Consul datacenter the client belongs to."
```
lkysow, line 173: 🍺

lkysow (approved):
Awesome work!

PR #68, opened 2021-11-16
Setup performance testing framework

erichaberkorn:
This shouldn't be merged.

- Setup a performance testing framework. See the README for more details.

## How I've tested this PR:
I've run a bunch of tests as described in the README.

## Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added


test/acceptance/perf/setup/service-group/main.tf
```diff
@@ -0,0 +1,132 @@
...
+  tls                            = true
+  consul_server_ca_cert_arn      = var.ca_cert_arn
+  gossip_key_secret_arn          = var.gossip_key_secret_arn
+  acls                           = true
+  acl_secret_name_prefix         = var.suffix
+  consul_client_token_secret_arn = var.consul_client_token_secret_arn
+  consul_ecs_image               = var.consul_ecs_image
+}
+
+# TODO These are annoying because it frequently blocks refreshes and deletes. Is there a better way to do this?
```
pglass, line 117: Would it help to create an "allow all" intention so that it isn't dependent on the task definitions?

test/acceptance/perf/setup/consul-server.tf
```diff
@@ -0,0 +1,244 @@
...
+  secret_id     = aws_secretsmanager_secret.bootstrap_token.id
+  secret_string = random_uuid.bootstrap_token.result
+}
+
+module "consul-server" {
+  // https://github.com/hashicorp/consul-global-scale-benchmark doesn't support ACLs so I forked it
+  source               = "github.com/erichaberkorn/consul-global-scale-benchmark/infrastructure/consul-server-ec2"
+  project              = local.name
+  vpc_id               = module.vpc.vpc_id
+  subnet_ids           = module.vpc.public_subnets
```
pglass, line 222: Since this is on public subnets, are the servers exposed to the public internet? And if so, are they restricted from the public internet somehow (by security group / ingress ip or etc)?

Just double checking that the Consul servers aren't subject to any extra/unknown load from "random hackers" trying to get into our EC2 instances.

test/acceptance/perf/setup/service-group/main.tf
```diff
@@ -0,0 +1,132 @@
...
+  desired_count   = var.server_instances_per_service_group
+  network_configuration {
+    subnets = var.private_subnets
+  }
+  propagate_tags         = "TASK_DEFINITION"
+  enable_execute_command = true
+
+  tags = var.tags
+  capacity_provider_strategy {
+    capacity_provider = "FARGATE_SPOT"
```
pglass, line 14: One concern with spot capacity is it can be pulled out from under us (interrupting our tasks / etc). 

Spot capacity seems useful to minimize cost while dev/testing the performance test itself, but when running the test proper, we should probably not use spot capacity to ensure consistency from test run to test run?

pglass (commented):
Looks good! Not fully reviewed, so just a few initial thoughts.

test/acceptance/perf/setup/service-group/main.tf
erichaberkorn: I moved it there. It's more permissive, but that doesn't matter for these tests.

test/acceptance/perf/setup/consul-server.tf
erichaberkorn: I moved the Consul servers to a private subnet

test/acceptance/perf/setup/service-group/main.tf
erichaberkorn: I'm going to try out spot instances to see if we can save some money. If they start causing problems, we can go back to regular instances.

lkysow (approved):
This looks really good!

PR #71, opened 2021-11-18
Allow passing in IAM roles to mesh-task

pglass:
* Add `task_role_arn` and `execution_role_arn` variables to allow users to pass in an IAM role. If not provide, they are created.
* No longer add SSM permissions by default

## How I've tested this PR:
Acceptance tests

👀 

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/modules/mesh-task/iam.tf
+++ b/modules/mesh-task/iam.tf
@@ -12,7 +12,7 @@ resource "aws_iam_role" "task" {
         Principal = {
           Service = "ecs-tasks.amazonaws.com"
         }
-      },
+      }
     ]
   })
 }
@@ -29,21 +29,19 @@ resource "aws_iam_role" "execution" {
   name  = "${var.family}-execution"
   path  = "/ecs/"
 
-  assume_role_policy = <<EOF
-{
-  "Version": "2012-10-17",
-  "Statement": [
-    {
-      "Sid": "",
-      "Effect": "Allow",
-      "Principal": {
-        "Service": "ecs-tasks.amazonaws.com"
-      },
-      "Action": "sts:AssumeRole"
-    }
-  ]
-}
-EOF
+  assume_role_policy = jsonencode({
+    Version = "2012-10-17"
+    Statement = [
+      {
+        Action = "sts:AssumeRole"
+        Effect = "Allow"
+        Principal = {
+          Service = "ecs-tasks.amazonaws.com"
+        }
+      }
+    ]
+  })
+
 }
 
 resource "aws_iam_policy" "execution" {
--- a/modules/mesh-task/iam.tf
+++ b/modules/mesh-task/iam.tf
@@ -15,34 +15,6 @@ resource "aws_iam_role" "task" {
       },
     ]
   })
-
-  tags = var.tags
-}
-
-resource "aws_iam_policy" "exec" {
-  name   = "${var.family}-execute-command"
-  path   = "/"
-  policy = <<EOF
-{
-  "Version": "2012-10-17",
-  "Statement": [
-    {
-      "Effect": "Allow",
-      "Action": [
-        "ssmmessages:CreateControlChannel",
-        "ssmmessages:CreateDataChannel",
-        "ssmmessages:OpenControlChannel",
-        "ssmmessages:OpenDataChannel"
-      ],
-      "Resource": [
-        "*"
-      ]
-    }
-  ]
-}
-EOF
-
-  tags = var.tags
 }
 
 resource "aws_iam_role_policy_attachment" "additional_task_policies" {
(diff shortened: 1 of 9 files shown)


modules/mesh-task/iam.tf
```diff
@@ -0,0 +1,130 @@
...
+  // to set a `count` field. So it errors.
+  //
+  // The workaround uses a splat to convert to 1-length or 0-length list.
+  //
+  // "If the value is anything other than a null value then the splat expression will transform
+  // it into a single-element list...If the value is null then the splat expression will return
+  // an empty tuple."
+  // https://www.terraform.io/docs/language/expressions/splat.html#single-values-as-lists
+  create_task_role      = length(var.task_role[*]) == 0
+  create_execution_role = length(var.execution_role[*]) == 0
```
pglass, line 16: So this seems to work.

Terraform complains when I try `var.task_role == null ? 1 : 0` though, and I'm not clear why that is. It seems like since TF is able to detect a null (i.e. variable not set) when using the splat, it should be able to do the same in other contexts.

modules/mesh-task/iam.tf
```diff
@@ -0,0 +1,130 @@
...
+}
```
lkysow, line 17: Could you also add

```hcl
  execution_role_id = local.create_execution_role ? aws_iam_role.execution[0].id : var.execution_role.id
  task_role_id = local.create_task_role ? aws_iam_role.task[0].id : var.task_role.id
```

So you don't duplicate this ternary in other places? Or does that not work.

modules/mesh-task/iam.tf
lkysow: 🧠 nice find

test/acceptance/tests/basic/terraform/basic-install/main.tf
```diff
@@ -266,6 +268,82 @@ module "test_server" {
   consul_client_token_secret_arn = var.secure ? module.acl_controller[0].client_token_secret_arn : ""
   acl_secret_name_prefix         = var.suffix
   consul_ecs_image               = var.consul_ecs_image
+
+  // Test passing both a role resource and role data source objects to make sure both
+  // have the necessary fields ("arn" and "id").
+  task_role                     = aws_iam_role.task
+  execution_role                = data.aws_iam_role.execution
+  additional_task_role_policies = [aws_iam_policy.exec.arn]
```
lkysow, line 276: why are you adding the exec policies to the task role?

test/acceptance/tests/basic/terraform/basic-install/main.tf
```diff
@@ -266,6 +268,82 @@ module "test_server" {
...
+      }
+    ]
+  })
+}
+
+// Test using an "existing" role name. Users will add a data source,
+// and pass the data source object into mesh-task.
+data "aws_iam_role" "execution" {
+  name = aws_iam_role.execution.name
 }
```
lkysow, line 347: Should we split this kind of test out into its own TF module? My thought is that it might make this basic-install test complicated when we try and excercise all our new behaviours in here.

lkysow (commented):
Looks good! Just a couple questions from me.

test/acceptance/tests/basic/terraform/basic-install/main.tf
pglass: Oh, this is bad naming. That's the policy to allow `aws execute-command`, which goes in the task role: https://github.com/hashicorp/terraform-aws-consul-ecs/blob/dd37985b6cb6608018772cd85f265a2585a1335c/modules/mesh-task/main.tf#L50-L87

test/acceptance/tests/basic/terraform/basic-install/main.tf
pglass: Yeah, I think I can move this bit into a test which runs a Terraform plan to validate this.

----

In general, we need to balance test runtime and test organization... Having quick tests is valuable, and Eric and I were already feeling pain from the current 30m test runtime (mostly from the HCP test).

We've so far favored test runtime... so we've updated the `basic-install` deployment to test a variety of things using only two mesh-tasks and a single Go test, which isn't great. We could improve that by adding additional mesh-tasks to test specific configurations, and by breaking validations up into Go sub-tests. That would help us test more cases with a single deployment while keeping the things somewhat organized.

But, that only goes so far too. Like testing an insecure vs secure config vs HCP servers need to be a separate deployment.

pglass: So, I only moved the bit where it passes the IAM role data source into a separate test, but I still test passing the iam role resources into mesh-task in basic-install.

I'm realizing now that you probably meant moving all of the "passing roles" test logic into a separate test (i.e. testing passing in a IAM role data source and IAM role resource)?

lkysow (approved):
🎉  thanks for doing the refactor!

PR #80, opened 2021-12-21
mesh-task: Support passing additional consul ecs config fields

pglass:
- Switch `var.upstreams` variable to camelCase, and update it to match the fields accepted by the consul-ecs config file.
- Add `var.consul_ecs_config` which supports only the fields of the consul-ecs config file that are not already covered by other Terraform variables
- Also, touched up some other variable descriptions to use Markdown formatting for the Terraform registry

## Checklist:
- [x] Depends on #79 
- [x] Tests added
- [x] CHANGELOG entry added 


modules/mesh-task/variables.tf
```diff
@@ -106,41 +106,66 @@ variable "envoy_image" {
...
+          key
+        )
+      ]
+    ]))
+  }
 }
 
 variable "checks" {
-  description = "A list of maps defining Consul checks for this service (https://www.consul.io/api-docs/agent/check#register-check)."
+  description = "A list of maps defining Consul checks for this service. See the Consul [Register Check documentation](https://www.consul.io/api-docs/agent/check#register-check)."
```
pglass, line 162: todo: Add validation and update description to match the consul-ecs config schema

modules/mesh-task/variables.tf
```diff
@@ -207,3 +232,115 @@ variable "application_shutdown_delay_seconds" {
   EOT
   default     = null
 }
+
+variable "consul_ecs_config" {
+  type        = any
```
pglass, line 237: I tried a [structural type](https://www.terraform.io/language/expressions/type-constraints#structural-types) to describe this, but that requires users pass all the nested fields here, even if they only want to specify a few. There is experimental support for optional fields and defaults in Terraform, but I didn't want to rely on that since it is not yet stable.

Instead, I use `type = any` and some validation to this variable to ensure users are passing the correct fields. It does not validate the types. I thought some immediate validation in Terraform is good feedback, and the full JSON Schema validation will occur at runtime.

Structural type example:

```hcl
type = object({
  service = object({
    enableTagOverride = bool
    weights = object({
      passing = number
      warning = number
    })
    namespace = string
  })
  proxy = object({
    meshGateway = object({
      mode = string
    })
    expose = object({
      checks = bool
      paths = list(object({
        listenerPort = number
        path = string
        localPathPort = number
        protocol = string
      }))
    })
  })
})
```

modules/mesh-task/variables.tf
```diff
@@ -207,3 +232,115 @@ variable "application_shutdown_delay_seconds" {
...
+  default     = {}
+  description = <<EOT
+  Additional configuration to pass to the consul-ecs binary for Consul service and sidecar proxy registration
+  requests.
+
+  This accepts a subset of the [consul-ecs config file](https://github.com/hashicorp/consul-ecs/blob/main/config/schema.json).
+  For the remainder of the consul-ecs config file contents, use the variables `upstreams`, `checks`, `consul_service_name`,
+  `consul_service_tags`, and `consul_service_meta`. In most cases, these separate variables should suffice.
```
pglass, line 245: This `consul_ecs_config` variable only includes the _additional_ fields we don't already support as dedicated Terraform variables. For example, we already have an `upstreams` variable, so I've excluded `service.proxy.upstreams` from the `consul_ecs_config` variable. Users would continue using the `upstreams` variable to pass in the upstreams.

This works. I personally wouldn't mind deprecating/removing some of the dedicated Terraform variables, in favor of passing them in as part of the config file. In particular, `consul_service_tags`, `consul_service_meta`, and `consul_service_name` could be removed and in favor of using this config variable to pass those. Thoughts?

modules/mesh-task/variables.tf
```diff
@@ -106,41 +106,66 @@ variable "envoy_image" {
...
-    object({
-      destination_name = string
-      local_bind_port  = number
-    })
-  )
+  type    = any
   default = []
+
+  validation {
+    error_message = "Upstream fields 'destinationName' and 'localBindPort' are required."
```
pglass, line 130: Note: upstream fields are now camelCase, which is a breaking change

modules/mesh-task/variables.tf
```diff
@@ -207,3 +232,115 @@ variable "application_shutdown_delay_seconds" {
...
+      contains(["service", "proxy"], key)
+    ])
+  }
+
+  validation {
+    error_message = "Only the 'enableTagOverride', 'weights', and 'namespace' fields are allowed in consul_ecs_config.service."
+    condition = alltrue([
+      for key in keys(lookup(var.consul_ecs_config, "service", {})) :
+      contains(["enableTagOverride", "weights", "namespace"], key)
+    ])
```
pglass, line 292: We _cannot_ use local variables in these conditions 😢

modules/mesh-task/variables.tf
erichaberkorn: IMO we should either go all in on either using individual variables for everything or the ``consul_ecs_config` variable for everything. Overall, I think using the `consul_ecs_config` variable for everything is better since it matches the interface for consul-ecs. This is a bunch larger breaking change though. What do you think?

modules/mesh-task/variables.tf
erichaberkorn: We could also look into using a JSON Schema Terraform provider to validate this data. I'm not certain if any of them are well supported and widely used though.

modules/mesh-task/variables.tf
```diff
@@ -106,41 +106,101 @@ variable "envoy_image" {
...
 
 variable "checks" {
-  description = "A list of maps defining Consul checks for this service (https://www.consul.io/api-docs/agent/check#register-check)."
-  type        = list(any)
+  description = "A list of maps defining Consul checks for this service. See the Consul [Register Check documentation](https://www.consul.io/api-docs/agent/check#register-check)."
+  type        = any
   default     = []
+
+  validation {
+    error_message = "Check fields must be one of 'checkId', 'name', 'scriptArgs', 'items', 'interval', 'timeout', 'ttl', 'http', 'header', 'method', 'body', 'tcp', 'status', 'notes', 'tlsServerName', 'tlsSkipVerify', 'grpc', 'grpcUseTls', 'aliasNode', 'aliasService', 'successBeforePassing', or 'failuresBeforeCritical'."
```
erichaberkorn, line 167: Nice work. Coming up with the exact required arguments must have been a ton of work

erichaberkorn (approved):
Nice work. I think we should come to consensus on [this](https://github.com/hashicorp/terraform-aws-consul-ecs/pull/80#discussion_r783267512) before the next release, but this looks much better to me.

modules/mesh-task/variables.tf
pglass: > We could also look into using a JSON Schema Terraform provider to validate this data.

Yeah, great idea. I was definitely interested in doing this. But there's only [one](https://registry.terraform.io/providers/xxxbobrxxx/jsonschema/latest) that I could find and it has not been updated. There's a slightly updated [fork](https://github.com/JeffAshton/terraform-provider-jsonschema) though, but it's still on Go 1.15 for example.

The other implication is users need to know to include any providers we use in their `required_providers` list, so I'm hesitant to add dependencies without a bit more consideration, and it didn't seem essential here.

I'd prefer we actually explore creating our own "helper" provider for things like validation and utilities. I don't think it'd be all that much work. And since it's all Golang, it would give us the ability to do some code sharing with our consul-ecs binary.

erichaberkorn: I agree completely 👍  We can definitely do this as enhancement later

modules/mesh-task/variables.tf
pglass: I'm not sure. The UX is matters a bit here. I was leaning toward keeping Terraform variables for commonly-used fields, and falling back on `consul_ecs_config` for less commonly-used fields. That way there's a bit less nesting, we can add nice TF documentation for those variables, and such.

@lkysow Thoughts?

erichaberkorn: If we go in that direction, we would probably want to consider which fields should have explicit variables rather than just relying on the ones that already have variables.

lkysow: I like offering the variables for things most people will interact with and the `consul_ecs_config` for power users or overrides.

pglass: > consider which fields should have explicit variables

Right. These are the variables we have currently from the consul-ecs config file:

| consul-ecs config  | terraform variable |
| ------------- | ------------- |
| `service.port`  | `port`  |
| `service.name`  | `consul_service_name`  |
| `service.tags` | `consul_service_tags` |
| `service.meta` | `consul_service_meta` |
| `service.checks` | `checks` |
| `proxy.upstreams` | `upstreams` | 

Do we want to change any of these? I suggested moving the name, tags, and meta into `consul_ecs_config`. But at least tags and meta were added based on user requests.

And we could add a TF variable for `namespace`?

lkysow: yeah consul_namespace would be fine

PR #82, opened 2022-01-12
Support passing Consul Agent configuration

pglass:
* Add a `consul_agent_configuration` variable to support passing an additional config file to the Consul agent.

## How I've tested this PR:
Acceptance tests, and also ran the example with additional Consul config

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/modules/mesh-task/templates/consul_client_command.tpl
+++ b/modules/mesh-task/templates/consul_client_command.tpl
@@ -2,36 +2,23 @@ cp /bin/consul /bin/consul-inject/consul
 
 ECS_IPV4=$(curl -s $ECS_CONTAINER_METADATA_URI_V4 | jq -r '.Networks[0].IPv4Addresses[0]')
 
-%{ if tls }
+%{ if tls ~}
 echo "$CONSUL_CACERT" > /tmp/consul-ca-cert.pem
-%{ endif }
+%{ endif ~}
+
+cat << EOF > /consul/agent-defaults.hcl
+${consul_agent_defaults_hcl}
+EOF
+
+%{ if consul_agent_configuration_hcl != null ~}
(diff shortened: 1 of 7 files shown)


modules/mesh-task/templates/consul_client_command.tpl
```diff
@@ -2,36 +2,23 @@ cp /bin/consul /bin/consul-inject/consul
...
 %{ endif ~}
-%{ if acls ~}
-  -hcl='acl {enabled = true, default_policy = "deny", down_policy = "async-cache"}' \
-  -hcl="acl { tokens { agent = \"$AGENT_TOKEN\" } }" \
+
+exec consul agent \
+    -data-dir /consul/data \
+    -config-file /consul/agent-defaults.hcl \
+%{ if consul_agent_configuration_hcl != null ~}
+    -config-file /consul/agent-extra.hcl
```
pglass, line 23: I was originally trying to also support Consul CLI flags. But after some careful review, all Consul flags have an equivalent config file option, with the exception of a few that don't matter (`-dev`, `-config-dir`, `-config-file`, `-config-format`, and `-hcl` are the only flags without a config file option).

This now uses only config files to pass options to Consul. And we rely on [Consul's config merging behavior](https://www.consul.io/docs/agent/options) when passing multiple files:

> Configuration specified later will be merged into configuration specified earlier. In most cases, "merge" means that the later version will override the earlier. In some cases, such as event handlers, merging appends the handlers to the existing configuration. The exact merging behavior is specified for each option below.

lkysow: yeah no need to support cli flags imo

lkysow (approved):

erichaberkorn (approved):

PR #87, opened 2022-02-14
add partition and namespace support

cthain:
- Document prerequisites for the acceptance tests in the README;
- Add a step to vet the local test execution environment prior to terraforming, in order to catch potential setup issues before the target environment is created;
- Add a test to verify upstream communication across namespaces is controlled correctly via intentions.
- Add a test to verify upstream communication across partitions and namespaces is controlled via `exported-services` and intentions.

## How I've tested this PR:
- Ran all the acceptance tests on ECS using HCP Consul and the latest `consul-ecs`;
- Ran the example in the `examples/admin-partitions` directory.

- :eyes: 
- Run the example and/or acceptance tests if you're feeling ambitious.

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added


test/acceptance/tests/hcp/terraform/ns/main.tf
```diff
@@ -0,0 +1,264 @@
...
+      }
+    ]
+    linuxParameters = {
+      initProcessEnabled = true
+    }
+  }]
+  retry_join = jsondecode(base64decode(hcp_consul_cluster.this.consul_config_file))["retry_join"]
+  upstreams = [
+    {
+      destinationName = "test_server_${var.suffix}"
```
pglass, line 156: In this upstream config for the client, you should need `destinationNamespace = local.test_server_ns` here. This routes the requests from the client service to the test server in that separate namespace.

pglass (commented):
Overall, great work! There's one important fix, which I commented about.

test/acceptance/tests/hcp/terraform/ns/main.tf
cthain: Good catch, thanks! I've updated per the comment.

cthain:
@pglass @erichaberkorn I've committed and pushed all my changes for the terraform repo related to the AP and NS support. Please have a look and let me know if you have any comments/suggestions; particularly with how I re-structured the HCP testing. Thanks!

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## Unreleased
+
+FEATURES
+* Added support for Consul Enterprise Admin Partitions and Namespaces.
```
cthain, line 4: Should be `Add support...`

test/acceptance/tests/hcp/hcp_test.go
```diff
@@ -1,124 +1,119 @@
...
 	"github.com/hashicorp/consul/sdk/testutil/retry"
+	"github.com/hashicorp/terraform-aws-consul-ecs/test/acceptance/framework/config"
 	"github.com/hashicorp/terraform-aws-consul-ecs/test/acceptance/framework/helpers"
 	"github.com/hashicorp/terraform-aws-consul-ecs/test/acceptance/framework/logger"
 	"github.com/stretchr/testify/require"
 )
 
+const setupDir = "../../setup-terraform"
+
+var timeouts = map[string]*retry.Timer{
```
erichaberkorn, line 23: It might be cleaner to define and use a struct since we know all of the keys. I might be missing something though.

test/acceptance/tests/hcp/hcp_test.go
```diff
@@ -1,124 +1,119 @@
...
-	logger.Log(t, "checking if services are registered")
-	retry.RunWith(&retry.Timer{Timeout: 6 * time.Minute, Wait: 20 * time.Second}, t, func(r *retry.R) {
-		services, _, err := consulClient.Catalog().Services(nil)
-		r.Check(err)
-		logger.Logf(t, "Consul services: %v", services)
-		require.Contains(r, services, serverServiceName)
-		require.Contains(r, services, clientServiceName)
+	logger.Log(t, "waiting for services to register with consul")
+	retry.RunWith(timeouts["registration"], t, func(r *retry.R) {
+		require.True(r, clientTask.Registered())
```
erichaberkorn, line 93: The code reads much better after this refactoring. Great work!

test/acceptance/tests/hcp/hcp_test.go
```diff
@@ -155,6 +150,402 @@ func TestHCP(t *testing.T) {
...
+func TestNamespaces(t *testing.T) {
+	// read the configuration from the setup-terraform dir.
+	var cfg HCPTestConfig
+	require.NoError(t, config.UnmarshalTF(setupDir, &cfg))
+
+	// generate input variables to the test terraform using the config.
+	ignoreVars := []string{"ecs_cluster_1_arn", "ecs_cluster_2_arn"}
+	tfVars := config.TFVars(cfg, ignoreVars...)
+
+	consulClient, consulState, err := consulClient(t, cfg.ConsulAddr, cfg.ConsulToken)
```
erichaberkorn, line 164: I think naming `consulState` something like `initialConsulState` would be more clear about its intention. This threw me off for a while.

test/acceptance/tests/hcp/hcp_test.go
```diff
@@ -155,6 +150,402 @@ func TestHCP(t *testing.T) {
...
+type NamespaceState struct {
+	Name        string
+	Tokens      map[string]struct{}
+	Policies    map[string]struct{}
+	Roles       map[string]struct{}
+	AuthMethods map[string]struct{}
+}
+
+func recordConsulState(t *testing.T, consul *api.Client) (ConsulState, error) {
+	// TODO: not sure how to handle OSS vs Enterprise cases.. these tests currently use Enterprise.
```
erichaberkorn, line 432: This is much better than before. So, thanks a lot for making the enterprise cases significantly better 😄

test/acceptance/tests/hcp/hcp_test.go
```diff
@@ -155,6 +150,402 @@ func TestHCP(t *testing.T) {
...
+		return nil, ConsulState{}, fmt.Errorf("failed to create Consul client: %w", err)
+	}
+	state, err := recordConsulState(t, client)
+	if err != nil {
+		return nil, ConsulState{}, fmt.Errorf("failed to establish Consul state: %w", err)
+	}
+	return client, state, err
+}
+
+type ConsulState struct {
```
erichaberkorn, line 414: I'm just curious, how does this deal with restoring the state of services?

examples/admin-partitions/README.md
```diff
@@ -0,0 +1,70 @@
+# Consul on ECS - Admin Partitions and Namespaces
+
+This folder provides an example of using the Admin Partition and Namespace features Consul Enterprise on ECS.
```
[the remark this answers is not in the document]
cthain: Agreed.

test/acceptance/framework/config/config.go
```diff
@@ -40,3 +48,90 @@ func (t TestConfig) TFVars(ignoreVars ...string) map[string]interface{} {
 	}
 	return vars
 }
+
+// UnmarshalTF populates the cfg struct with the Terraform outputs
+// from the given tfDir directory. The cfg arg must be a pointer to
+// a value that can be populated by json.Unmarshal based on the output
+// of the `terraform output -json` command.
+func UnmarshalTF(tfDir string, cfg interface{}) error {
```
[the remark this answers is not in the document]
cthain: Yes currently it is only used by the HCP tests. I can be explicit about the type of the input arg, in which case it may also be best to move it into the `hcp` test package. I made this func generic and included it in the `config` package with the intention of eventually refactoring the `suite` package to use it.. I didn't want to make those changes under this PR because I didn't want to make non-functional changes to tests that I didn't change.

Let me know what you think.

test/acceptance/tests/hcp/hcp_test.go
cthain: Great suggestion!

test/acceptance/tests/hcp/hcp_test.go
cthain: Thanks :)

test/acceptance/tests/hcp/hcp_test.go
cthain: Agreed.. Sorry for the confusion.

test/acceptance/tests/hcp/hcp_test.go
cthain: Currently it only handles ACL state that may be changed by the ACL controller. Do you think it is worth changing the name of the struct to something like `ConsulACLState`?

test/acceptance/tests/hcp/hcp_test.go
```diff
@@ -155,6 +150,402 @@ func TestHCP(t *testing.T) {
...
+			}
+
+			partState.Namespaces[ns.Name] = nsState
+		}
+		state.Partitions[part.Name] = partState
+	}
+	return state, nil
+}
+
+func restoreConsulState(t *testing.T, consul *api.Client, state ConsulState) error {
```
pglass, line 481: I think this is fine, but to simplify, is there a reason we shouldn't delete everything in the HCP cluster? We could clean out the HCP cluster to ensure a clean slate at the beginning of each test case.

I guess we couldn't parallelize easily if we do that (although, we maybe can't parallelize these easily either when the test cases share a server cluster?).

pglass (approved):
This is really great!

(In general, I'd encourage more smaller PRs next time, but I'm guilty of large single PRs too so 👍 )

CHANGELOG.md
cthain: Agreed.. Not really too much of a chicken/egg in this case because I already had the PR created.

test/acceptance/tests/hcp/hcp_test.go
cthain: My pleasure :)

test/acceptance/tests/hcp/hcp_test.go
```diff
@@ -155,6 +150,402 @@ func TestHCP(t *testing.T) {
...
+	var cfg HCPTestConfig
+	require.NoError(t, config.UnmarshalTF(setupDir, &cfg))
+
+	// generate input variables to the test terraform using the config.
+	ignoreVars := []string{"ecs_cluster_1_arn", "ecs_cluster_2_arn"}
+	tfVars := config.TFVars(cfg, ignoreVars...)
+
+	consulClient, consulState, err := consulClient(t, cfg.ConsulAddr, cfg.ConsulToken)
+	require.NoError(t, err)
+	defer func() {
```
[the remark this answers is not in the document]
cthain: Agreed.

test/acceptance/tests/hcp/hcp_test.go
```diff
@@ -1,124 +1,119 @@
...
+	ECSCluster2ARN          string      `json:"ecs_cluster_2_arn"`
+	ConsulAddr              string      `json:"consul_public_endpoint_url"`
+	ConsulPrivateAddr       string      `json:"consul_private_endpoint_url"`
+	RetryJoin               interface{} `json:"retry_join"`
+	BootstrapTokenSecretARN string      `json:"bootstrap_token_secret_arn"`
+	GossipKeySecretARN      string      `json:"gossip_key_secret_arn"`
+	ConsulCASecretARN       string      `json:"consul_ca_cert_secret_arn"`
+
+	// Fields that are excluded from the Terraform variables have no JSON tags.
+	ConsulToken string
```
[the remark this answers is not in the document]
cthain: > Not that it really matters here, but it will parse a "ConsulToken" field (case insensitively), if present.

Good catch!

In the end we actually need the `token` field to be serialized/deserialized because the tests need it to authenticate the `consulClient`.. this is why the HCP tests failed in the CI pipeline.

test/acceptance/framework/helpers/task.go
```diff
@@ -0,0 +1,134 @@
...
+// MeshTask represents a Consul ECS mesh task and provides utility functions for interacting with it.
+type MeshTask struct {
+	ConsulClient *api.Client
+	Name         string
+	Partition    string
+	Namespace    string
+	Region       string
+	ClusterARN   string
+	taskARN      string
+}
```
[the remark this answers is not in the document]
cthain: Good call. I combined your suggestion with Eric's and I added a `NewMeshTask` constructor that uses `t.NotNil/Empty` to validate the inputs up front.

examples/admin-partitions/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+![Admin Partitions Example](context.png)
```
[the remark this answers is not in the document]
cthain: Agreed.

examples/admin-partitions/terraform/aws_ecs.tf
```diff
@@ -0,0 +1,106 @@
...
+
+  name = local.ecs_name
+  cidr = "10.0.0.0/16"
+  // The NAT gateway limit is per AZ. With `single_nat_gateway = true`, the NAT gateway is created
+  // in the first public subnet. Shuffling AZs helps spread NAT gateways across AZs to help with this.
+  azs = [
+    // Silly, but avoids this error: `"count" value depends on resource attributes that cannot be determined until apply`
+    random_shuffle.azs.result[0],
+    random_shuffle.azs.result[1],
+    random_shuffle.azs.result[2],
```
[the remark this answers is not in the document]
cthain: Good point and thanks for the clarification. I've update the example.

test/acceptance/framework/helpers/task.go
```diff
@@ -0,0 +1,134 @@
...
+
+// Registered indicates if the service for the task is registered in Consul.
+func (task *MeshTask) Registered() bool {
+	if task.ConsulClient == nil {
+		panic("MeshTask.Registered() called with nil Consul client")
```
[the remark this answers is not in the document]
cthain: I combined your suggestion with Paul's and created a `NewMeshTask` constructor that takes in `*testing.T` and uses `require.NotNil/Empty` to vet the input args up front.

test/acceptance/framework/config/config.go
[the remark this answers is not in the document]
cthain: For now I've moved the HCP-specific config into a new `config.go` file in the `hcp` package. It explicitly uses the `HCPTestConfig` struct now.

examples/admin-partitions/ap-example
```diff
@@ -0,0 +1,74 @@
...
+
+# wait for services to be registered in Consul
+waitfor ${CLIENT_AP} ${CLIENT_NS} ${CLIENT_NAME}
+echo ""
+
+waitfor ${SERVER_AP} ${SERVER_NS} ${SERVER_NAME}
+echo ""
+
+# create exported-services config entry to expose the example_server in partition 2 to the services in partition 1
+echo "creating exported-services config entry"
```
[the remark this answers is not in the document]
cthain: Both are good suggestions. Long term I like the idea of using the Terraform consul provider to do this since I think that is the best practicable example we can provide.

The example, as it currently is, is a bit contrived and probably not how things would be done in practice. I did this to test that the ACL controller was doing specific things like creating the partition and namespace(s) for the tasks if they don't already exist. But those types of checks belong in the acceptance tests (which they do now), not an example.

> Actually, it looks like intentions are managed via config entries as well (TIL)

Yes, I learned that recently too and I updated the HCP acceptance tests to use config-entry-style intentions.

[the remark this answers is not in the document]
cthain: The Consul TF provider does not support admin partitions. I updated the `hcp_consul.tf` file with the proposed updates and have commented them out with a TODO for now. I have added a task to the backlog to update this once the TF provider supports admin partitions.

I am going to leave the script as is (using raw HTTP requests) for now since I think the TF provider is the way to go and I don't want to update this thrice.

test/acceptance/tests/hcp/hcp_test.go
cthain: I've left this as is for now. I've create a task in the backlog to revisit the acceptance test performance.

test/acceptance/tests/hcp/hcp_test.go
cthain: Parallel == Hard :laughing: 

I've left this as is for now. I've create a task in the backlog to revisit the acceptance test performance.

test/acceptance/tests/hcp/hcp_test.go
erichaberkorn: Yeah, that is even better @pglass

PR #90, opened 2022-03-29
Improve Terraform Registry documentation

pglass:
- Remove markdown that doesn't render for modules in the Terraform registry.
- Avoid `default = null` for optional variables.

## How I've tested this PR:

- [Preview with a personal test module](https://registry.terraform.io/modules/pglass/testmodule/pglass/latest/submodules/mesh-task?tab=inputs). 
  - I copied the `variables.tf` files to a [test repo](https://github.com/pglass/terraform-pglass-testmodule) in order to preview the docs in the TF registry.
- Acceptance tests

- Double check logic changes for non-null default values

## Checklist:
- [x] Tests added
- [ ] ~CHANGELOG entry added~ n/a

### DIFF
--- a/modules/mesh-task/iam.tf
+++ b/modules/mesh-task/iam.tf
@@ -12,8 +12,8 @@ locals {
   // it into a single-element list...If the value is null then the splat expression will return
   // an empty tuple."
   // https://www.terraform.io/docs/language/expressions/splat.html#single-values-as-lists
-  create_task_role      = length(var.task_role[*]) == 0
-  create_execution_role = length(var.execution_role[*]) == 0
+  create_task_role      = length(var.task_role[*]) == 0 ? true : length(var.task_role.id[*]) == 0
+  create_execution_role = length(var.execution_role[*]) == 0 ? true : length(var.execution_role.id[*]) == 0
 
   execution_role_id = local.create_execution_role ? aws_iam_role.execution[0].id : var.execution_role.id
   task_role_id      = local.create_task_role ? aws_iam_role.task[0].id : var.task_role.id
--- a/modules/mesh-task/variables.tf
+++ b/modules/mesh-task/variables.tf
@@ -24,13 +24,13 @@ variable "consul_service_meta" {
 variable "consul_namespace" {
   description = "The Consul namespace to use to register this service [Consul Enterprise]."
   type        = string
-  default     = null
+  default     = ""
 }
 
 variable "consul_partition" {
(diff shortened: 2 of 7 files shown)


modules/mesh-task/iam.tf
```diff
@@ -12,8 +12,8 @@ locals {
   // it into a single-element list...If the value is null then the splat expression will return
   // an empty tuple."
   // https://www.terraform.io/docs/language/expressions/splat.html#single-values-as-lists
-  create_task_role      = length(var.task_role[*]) == 0
-  create_execution_role = length(var.execution_role[*]) == 0
+  create_task_role      = length(var.task_role[*]) == 0 ? true : length(var.task_role.id[*]) == 0
+  create_execution_role = length(var.execution_role[*]) == 0 ? true : length(var.execution_role.id[*]) == 0
```
pglass, line 16: Terraform conditionals do not short circuit, so using `length(var.task_role[*]) == 0 || length(var.task_role.id[*]) == 0` will error when `var.task_role = null`. Hence, the workaround.

modules/mesh-task/variables.tf
```diff
@@ -275,64 +281,32 @@ variable "consul_datacenter" {
...
-            listenerPort = 1234
-            path = "/path"
-            localPathPort = 2345
-            protocol = "http"
-          }
-        ]
-      }
-    }
-  }
-  ```
```
pglass, line 335: This condenses to a single line in the TF registry, so I opted to just remove the example. These are infrequently used options anyway so I doubt it's that helpful.

test/acceptance/tests/basic/basic_test.go
```diff
@@ -128,6 +128,64 @@ func TestPassingExistingRoles(t *testing.T) {
...
+			if c.expEntrypoint {
+				// Look for app-entrypoint in the Terraform diff.
+				regex := strings.Join([]string{
+					`\+ entryPoint  = \[`,
+					`  \+ "/consul/consul-ecs",`,
+					`  \+ "app-entrypoint",`,
+					`  \+ "-shutdown-delay",`,
+					`  \+ "\d+s",`, // e.g. "2s", "10s", etc
+					`\]`,
+				}, `\s+`)
```
pglass, line 179: ew, regex. But this type of thing is useful for quick checks of the `terraform plan` output, without running a full `terraform apply` (which would need to at least create a task definition, so it would take longer).

PR #95, opened 2022-04-06
Correct secret name when admin partitions are enabled

cthain:
- Set secret name correctly when partition and namespace are present.
- Add validation for partition and namespace variables.
- Add acceptance tests for updates.

## How I've tested this PR:
- Ran the acceptance tests

:eyes: 

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added

### DIFF
--- a/modules/mesh-task/validation.tf
+++ b/modules/mesh-task/validation.tf
@@ -4,4 +4,5 @@ locals {
   require_client_token_if_acls_enabled       = (var.acls && var.consul_client_token_secret_arn == "") ? file("ERROR: consul_client_token_secret_arn must be set if acls is true") : null
   require_secret_name_prefix_if_acls_enabled = (var.acls && var.acl_secret_name_prefix == "") ? file("ERROR: acl_secret_name_prefix must be set if acls is true") : null
   require_namespace_if_partition_is_set      = (var.consul_partition != "" && var.consul_namespace == "") ? file("ERROR: consul_namespace must be set if consul_partition is set") : null
+  require_partition_if_namespace_is_set      = (var.consul_namespace != "" && var.consul_partition == "") ? file("ERROR: consul_partition must be set if consul_namespace is set") : null
 }
--- a/modules/mesh-task/validation.tf
+++ b/modules/mesh-task/validation.tf
@@ -3,4 +3,5 @@ locals {
   require_ca_cert_if_tls_enabled             = (var.tls && var.consul_server_ca_cert_arn == "") ? file("ERROR: consul_server_ca_cert_arn must be set if tls is true") : null
   require_client_token_if_acls_enabled       = (var.acls && var.consul_client_token_secret_arn == "") ? file("ERROR: consul_client_token_secret_arn must be set if acls is true") : null
   require_secret_name_prefix_if_acls_enabled = (var.acls && var.acl_secret_name_prefix == "") ? file("ERROR: acl_secret_name_prefix must be set if acls is true") : null
+  require_namespace_if_partition_is_set      = (var.consul_partition != "" && var.consul_namespace == "") ? file("ERROR: consul_namespace must be set if consul_partition is set") : null
 }
--- a/test/acceptance/tests/basic/basic_test.go
+++ b/test/acceptance/tests/basic/basic_test.go
@@ -352,10 +352,8 @@ func TestValidation_ConsulEcsConfigVariable(t *testing.T) {
 	}
 }
 
-// Test the validation that if a partition is provided, a namespace must also be provided.
-func TestValidation_NamespaceIsRequiredIfPartitionIsSet(t *testing.T) {
-	t.Parallel()
-
+// Test the validation that both partition and namespace must be provided or neither.
+func TestValidation_NamespaceAndPartitionRequired(t *testing.T) {
 	cases := map[string]struct {
 		partition string
 		namespace string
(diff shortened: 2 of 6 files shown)


modules/mesh-task/validation.tf
```diff
@@ -3,4 +3,5 @@ locals {
   require_ca_cert_if_tls_enabled             = (var.tls && var.consul_server_ca_cert_arn == "") ? file("ERROR: consul_server_ca_cert_arn must be set if tls is true") : null
   require_client_token_if_acls_enabled       = (var.acls && var.consul_client_token_secret_arn == "") ? file("ERROR: consul_client_token_secret_arn must be set if acls is true") : null
   require_secret_name_prefix_if_acls_enabled = (var.acls && var.acl_secret_name_prefix == "") ? file("ERROR: acl_secret_name_prefix must be set if acls is true") : null
+  require_namespace_if_partition_is_set      = (var.consul_partition != "" && var.consul_namespace == "") ? file("ERROR: consul_namespace must be set if consul_partition is set") : null
```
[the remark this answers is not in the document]
lkysow: or are we using the fact that it's not an empty string to know that it's consul ent?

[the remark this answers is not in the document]
cthain: because we need to support OSS as well as enterprise and you can't use any partition/namespace values with OSS as it causes the Consul agent to return errors.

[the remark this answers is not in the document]
cthain: I was chatting with @erichaberkorn about this today. I think there is a bit more work that we can do in terms of improving the UX around AP/NS support in the OSS/Enterprise cases. There are a few permutations that we could handle more gracefully in terms of defaulting partition/namespace values. I think the priority for now is to get this patch out because without it things are broken.

test/acceptance/tests/basic/basic_test.go
```diff
@@ -352,217 +352,306 @@ func TestValidation_ConsulEcsConfigVariable(t *testing.T) {
...
+		},
+		"without partition, with namespace": {
+			partition: "",
+			namespace: "default",
+			errMsg:    "",
+		},
+	}
+
+	for name, c := range cases {
+		t.Run(name, func(t *testing.T) {
```
pglass, line 387: non-blocking: For plan-only tests, I had found it's a touch faster to do the `terraform init` once outside the loop, and `t.Parallel()` here within the loop.

ex: https://github.com/hashicorp/terraform-aws-consul-ecs/blob/b8444236a962711e6070bd420723b2d07b99d8e6/test/acceptance/tests/basic/basic_test.go#L154-L157

test/acceptance/tests/basic/basic_test.go
```diff
@@ -352,217 +352,306 @@ func TestValidation_ConsulEcsConfigVariable(t *testing.T) {
...
+			randomSuffix := strings.ToLower(random.UniqueId())
+			tfVars := suite.Config().TFVars("route_table_ids")
+			tfVars["secure"] = secure
+			tfVars["suffix"] = randomSuffix
+			clientServiceName := "test_client"
+
+			serverServiceName := "test_server"
+			if !secure {
+				// This uses the explicitly passed service name rather than the task's family name.
+				serverServiceName = "custom_test_server"
```
pglass, line 424: Why is this service name different for insecure vs secure?

test/acceptance/tests/basic/terraform/basic-install/main.tf
```diff
@@ -281,10 +281,6 @@ module "test_server" {
   acl_secret_name_prefix         = var.suffix
   consul_ecs_image               = var.consul_ecs_image
 
-  // Test passing both a role resource and role data source objects to make sure both
-  // have the necessary fields ("arn" and "id").
-  task_role                     = aws_iam_role.task
-  execution_role                = aws_iam_role.execution
```
pglass, line 287: Why remove these?

modules/mesh-task/validation.tf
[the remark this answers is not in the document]
cthain: > or are we using the fact that it's not an empty string to know that it's consul ent?

Yes, that's correct.

test/acceptance/tests/basic/basic_test.go
cthain: The "insecure" case had it set to an explicit value and I thought it would be more complete to test with different permutations for service name. I can set them both to `custom_test_server` if that makes more sense.

pglass: 👍 That makes sense.

(Just wanted to check there was not some odd reason this was required to get the secure case passing.)

test/acceptance/tests/basic/terraform/basic-install/main.tf
cthain: They were causing the `secure` case to fail on the `terraform apply`. I'm trying to figure out exactly why. I wanted to get the rest of the PR out for review while I worked on figuring this out.

pglass: Oh, right. Because the execution role needs permission to fetch AWS Secrets Manager secrets in the secure case, and those perms aren't set by basic-install.

I think it's fine to leave this for a follow up, as long as we don't lose track of it.

pglass: Well, wait, but mesh-task should create a policy for those additional permissions:

https://github.com/hashicorp/terraform-aws-consul-ecs/blob/b8444236a962711e6070bd420723b2d07b99d8e6/modules/mesh-task/iam.tf#L71 

So actually I'm still not sure why it would fail.

test/acceptance/tests/basic/basic_test.go
cthain: I swapped the `serverServiceName` for the secure and insecure case and the test failed. The `secret_name` is using `var.family` but it should be using `local.service_name`. So the secure case was only passing coincidentally because `test_server` is the same as the `family`.

test/acceptance/tests/basic/basic_test.go
```diff
@@ -352,217 +352,306 @@ func TestValidation_ConsulEcsConfigVariable(t *testing.T) {
...
+			partition: "default",
+			namespace: "default",
+			errMsg:    "",
+		},
+		"with partition, without namespace": {
+			partition: "default",
+			namespace: "",
+			errMsg:    "ERROR: consul_namespace must be set if consul_partition is set",
+		},
+		"without partition, with namespace": {
```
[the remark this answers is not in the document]
cthain: Agreed. I'll add the validation for requiring partition when namespace is set.

test/acceptance/tests/basic/terraform/basic-install/main.tf
cthain: It fails due to a difference in the way that splat expressions are handled between Terraform v1.0 and v1.1. The CI pipeline uses v1.0 and I was using v1.1 locally. We will leave the test the as proposed in the PR for now and fix this issue in a future PR.

test/acceptance/tests/basic/basic_test.go
cthain: I've updated the test as suggested.

test/acceptance/tests/basic/basic_test.go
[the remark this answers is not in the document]
cthain: Added the validation and updated the test case as suggested.

pglass (approved):

PR #96, opened 2022-04-07
dev-server: Support Consul enterprise license

pglass:
- Add `consul_license` input variable to specify an enterprise license key. This is placed in a Secrets Manager secret that is the set as the `CONSUL_LICENSE` variable for the dev-server
- Add `wait_for_steady_state` input variable, which is passed through to the ECS service
- Add `bootstrap_token_id` output variable

## How I've tested this PR:
Manually

👀 

## Checklist:
- [ ] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/modules/dev-server/main.tf
+++ b/modules/dev-server/main.tf
@@ -6,7 +6,7 @@ locals {
     container_port   = 8500
   }] : []
 
-  consul_license_enabled = var.consul_license != ""
+  consul_enterprise_enabled = var.consul_license != ""
 }
 
 resource "tls_private_key" "ca" {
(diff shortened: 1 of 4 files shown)


modules/dev-server/main.tf
```diff
@@ -5,6 +5,8 @@ locals {
     container_name   = "consul-server"
     container_port   = 8500
   }] : []
+
+  consul_license_enabled = var.consul_license != ""
```
cthain, line 9: Would it make sense to call this `consul_enterprise` or `consul_enterprise_enabled`? I'm not sure if this will get re-used anywhere else in the future so maybe it's not worth it.

cthain (approved):
Looks great, nice work Paul!

PR #100, opened 2022-04-25
Support the auth method

pglass:

This is an incremental change which flags off the login functionality until the auth method is fully supported. By default, the old mechanism for tokens is used (secrets in Secrets Manager). The `client_token_auth_method_name`, `service_token_auth_method_name`, and `consul_http_addr` variables are required to be set to use the new mechanism which logs in to auth method. Later, we'll switch over to the auth method as the default for ACL tokens.

- The server CA cert file is now placed on the shared volume, `/consul/consul-ca-cert.pem`, so that mesh-init has access to the file for its `consul login`
- New input variables: `client_token_auth_method_name`, `service_token_auth_method_name`, and `consul_http_addr`
- The IAM task role is configured with an `iam:GetRole` permission for itself
- The acl-controller has a new `iam_role_path` variable which, if set, is passed as `-iam-role-path` to the `acl-controller` command

## How I've tested this PR:
Acceptance tests.
- The existing tests should still pass
- A new test, `TestBasic/secure:_true,authMethod:_true`, tests the auth method

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,7 +99,7 @@ jobs:
       # Restore go module cache if there is one
       - restore_cache:
           keys:
-            - consul-ecs-acceptance-modcache-v1-{{ checksum "test/acceptance/go.mod" }}
+            - consul-ecs-acceptance-modcache-v2-{{ checksum "test/acceptance/go.mod" }}
 
       - run: mkdir -p $TEST_RESULTS
 
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ orbs:
 executors:
   consul-ecs-test:
     docker:
-      - image: docker.mirror.hashicorp.services/hashicorpdev/consul-ecs-test:0.3.0
+      - image: docker.mirror.hashicorp.services/hashicorpdev/consul-ecs-test:0.3.1
     environment:
       TEST_RESULTS: &TEST_RESULTS /tmp/test-results # path to where test results are saved
 
@@ -17,7 +17,7 @@ jobs:
       # Restore go module cache if there is one
       - restore_cache:
           keys:
-            - consul-ecs-acceptance-modcache-v1-{{ checksum "test/acceptance/go.mod" }}
+            - consul-ecs-acceptance-modcache-v2-{{ checksum "test/acceptance/go.mod" }}
 
       - run:
           name: go mod download
@@ -26,9 +26,9 @@ jobs:
 
       # Save go module cache if the go.mod file has changed
       - save_cache:
-          key: consul-ecs-acceptance-modcache-v1-{{ checksum "test/acceptance/go.mod" }}
+          key: consul-ecs-acceptance-modcache-v2-{{ checksum "test/acceptance/go.mod" }}
           paths:
-            - "/go/pkg/mod"
+            - "/home/circleci/go/pkg/mod"
 
       # check go fmt output because it does not report non-zero when there are fmt changes
       - run:
@@ -108,7 +108,7 @@ jobs:
           working_directory: test/acceptance/tests
           no_output_timeout: 1h
           command: |
-            gotestsum --junitfile "$TEST_RESULTS/gotestsum-report.xml" -- ./... -p 1 -timeout 1h -v -failfast
+            gotestsum --junitfile "$TEST_RESULTS/gotestsum-report.xml" --format standard-verbose -- ./... -p 1 -timeout 1h -v -failfast
 
       - store_test_results:
           path: *TEST_RESULTS
--- a/modules/mesh-task/iam.tf
+++ b/modules/mesh-task/iam.tf
@@ -47,7 +47,8 @@ resource "aws_iam_role" "task" {
   }
 }
 
-// If acls enabled, we need the task role configure for the auth method.
+// If acls are enabled, the task role must be configured with an `iam:GetRole` permission
+// to fetch itself, in order to be compatbile with the auth method.
 resource "aws_iam_policy" "task" {
   count       = var.acls ? 1 : 0
   name        = "${var.family}-task"
--- a/modules/mesh-task/iam.tf
+++ b/modules/mesh-task/iam.tf
@@ -25,6 +25,7 @@ locals {
 // Create the task role
 resource "aws_iam_role" "task" {
   count = local.create_task_role ? 1 : 0
+  path  = "/ecs/"
 
   name = "${var.family}-task"
   assume_role_policy = jsonencode({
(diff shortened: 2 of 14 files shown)


test/acceptance/tests/basic/terraform/basic-install/main.tf
```diff
@@ -49,7 +49,7 @@ variable "launch_type" {
 variable "consul_ecs_image" {
   description = "Consul ECS image to use."
   type        = string
-  default     = "docker.mirror.hashicorp.services/hashicorpdev/consul-ecs:latest"
+  default     = "docker.mirror.hashicorp.services/hashicorpdev/consul-ecs:889076e"
```
pglass, line 52: I'll revert this once https://github.com/hashicorp/consul-ecs/pull/74 is merged

test/acceptance/tests/basic/terraform/basic-install/main.tf
```diff
@@ -49,7 +56,7 @@ variable "launch_type" {
 variable "consul_ecs_image" {
   description = "Consul ECS image to use."
   type        = string
-  default     = "docker.mirror.hashicorp.services/hashicorpdev/consul-ecs:latest"
+  default     = "docker.mirror.hashicorp.services/hashicorpdev/consul-ecs:7dcd59b"
```
pglass, line 59: Revert this once https://github.com/hashicorp/consul-ecs/pull/71 is merged

modules/mesh-task/iam.tf
```diff
@@ -39,6 +40,43 @@ resource "aws_iam_role" "task" {
...
     ]
   })
+
+  tags = {
+    "consul.hashicorp.com.service-name" = local.service_name
+    "consul.hashicorp.com.namespace"    = var.consul_namespace
+  }
+}
+
+// If acls enabled, we need the task role configure for the auth method.
```
cthain, line 50: sp: configure -> configured

modules/mesh-task/templates/consul_client_command.tpl
```diff
@@ -3,7 +3,32 @@ cp /bin/consul /bin/consul-inject/consul
...
+      -aws-auto-bearer-token -aws-include-entity \
+      -token-sink-file /consul/token.txt
+}
+
+while ! login; do
+    sleep 2
+done
+
+# This is an env var which is interpolated into the agent-defaults.hcl
+export AGENT_TOKEN=$(cat /consul/token.txt)
```
cthain, line 31: This env var will be set once after the login script completes successfully.. The docs for the [`-token-sink-file`](https://www.consul.io/commands/login#token-sink-file) option seem to suggest that the token secret ID in this file may be updated. I'm not sure under what circumstances this would happen, or if it even would.

I think what is here is fine, I had a look through the code and it appears that the file is only ever written to once, after a successful login.. unless I'm missing something: https://github.com/hashicorp/consul/blob/main/command/login/login.go#L159

Curious about your thoughts or if you checked into this.

test/acceptance/tests/basic/terraform/basic-install/main.tf
```diff
@@ -299,6 +320,14 @@ resource "aws_iam_role" "task" {
       },
     ]
   })
+
+  // TODO: When using the auth method, mesh-task should add these tags if the task role
+  // is passed in.
+  tags = {
+    "consul.hashicorp.com.service-name" = "${var.server_service_name}_${var.suffix}"
+    // "consul.hashicorp.com.namespace"    = ""
```
cthain, line 328: why is namespace commented out?

cthain (approved):
I've finished my review, looks good Paul.

modules/mesh-task/templates/consul_client_command.tpl
pglass: Right, the Consul client/agent doesn't rewrite that token file. Only the `consul login` command would update that file, if it already exists, in order to make it easy to refresh tokens. This would matter if we want to support token expiration (`consul acl auth-method create -max-token-ttl <duration> ...`) but we aren't supporting token expiration for now.

.circleci/config.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       # Restore go module cache if there is one
       - restore_cache:
           keys:
-            - consul-ecs-acceptance-modcache-v1-{{ checksum "test/acceptance/go.mod" }}
+            - consul-ecs-acceptance-modcache-v2-{{ checksum "test/acceptance/go.mod" }}
```
pglass, line 20: We had change the consul-ecs-test image to use the newer CircleCI image (`cimg/go`) which changes the go path, hence these other changes.

test/acceptance/tests/basic/terraform/basic-install/main.tf
```diff
@@ -299,6 +323,14 @@ resource "aws_iam_role" "task" {
       },
     ]
   })
+
+  // TODO: When using the auth method, mesh-task should add these tags if the task role
+  // is passed in.
```
pglass, line 328: It doesn't look like we can add individual tags to an existing role that is passed in.

Terraform does support [managing individual tags for certain resources](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/resource-tagging#managing-individual-resource-tags), but there is no resource for IAM role tags, unfortunately.

For now, if a user passes their own role, they will need to add these tags to the role on their side (if they have ACLs enabled and are using the auth method).

modules/mesh-task/templates/consul_client_command.tpl
```diff
@@ -1,9 +1,37 @@
 cp /bin/consul /bin/consul-inject/consul
 
-ECS_IPV4=$(curl -s $ECS_CONTAINER_METADATA_URI_V4 | jq -r '.Networks[0].IPv4Addresses[0]')
+ECS_TASK_META=$(curl -s $ECS_CONTAINER_METADATA_URI_V4)
+ECS_IPV4=$(echo "$ECS_TASK_META" | jq -r '.Networks[0].IPv4Addresses[0]')
+TASK_REGION=$(echo "$ECS_TASK_META" | jq -r .ContainerARN | cut -d':' -f4)
```
pglass, line 5: On EC2, the AWS region is not set in the environment, so we parse it from the task metadata here.

.circleci/config.yml
pglass: TODO: I missed updating a spot below in this file where the modcache is restore (will update that after the current acceptance test run finishes)

.circleci/config.yml
```diff
@@ -108,7 +108,7 @@ jobs:
           working_directory: test/acceptance/tests
           no_output_timeout: 1h
           command: |
-            gotestsum --junitfile "$TEST_RESULTS/gotestsum-report.xml" -- ./... -p 1 -timeout 1h -v -failfast
+            gotestsum --junitfile "$TEST_RESULTS/gotestsum-report.xml" --format standard-verbose -- ./... -p 1 -timeout 1h -v -failfast
```
pglass, line 111: This gets us streaming output in CI, so we can see acceptance test progress.

PR #102, opened 2022-04-28
Mitigate against token replication lag

pglass:
Copy the approach from https://github.com/hashicorp/consul-k8s/pull/887 for dealing with inherent token replication lag.

See: https://github.com/hashicorp/consul-ecs/pull/79

## How I've tested this PR:
Acceptance tests

👀 

## Checklist:
- [ ] Tests added
- [ ] ~CHANGELOG entry added~ n/a

### DIFF
--- a/modules/mesh-task/templates/consul_client_command.tpl
+++ b/modules/mesh-task/templates/consul_client_command.tpl
@@ -26,9 +26,36 @@ login() {
       -token-sink-file /consul/client-token
 }
 
+read_token_stale() {
+    consul acl token read -http-addr ${ consul_http_addr } \
+    %{ if tls ~}
+      -ca-file /consul/consul-ca-cert.pem \
+    %{ endif ~}
+      -stale -self -token-file /consul/client-token
+}
+
+# Retry in order to login successfully.
 while ! login; do
     sleep 2
 done
+
+# Wait for raft replication to hopefully occur. Without this, an "ACL not found" may be cached for a while.
+# Technically, the problem could still occur but this should handle most cases.
+# This waits at most 2s (20 attempts with 0.1s sleep)
+COUNT=20
+while [ "$COUNT" -gt "0" ]; do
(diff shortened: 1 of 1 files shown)


modules/mesh-task/templates/consul_client_command.tpl
```diff
@@ -26,10 +26,37 @@ login() {
       -token-sink-file /consul/client-token
 }
 
+read_token_stale() {
+    consul acl token read -http-addr ${ consul_http_addr } \
+    %{ if tls ~}
+      -ca-file /consul/consul-ca-cert.pem \
+    %{ endif ~}
+      -stale -self -token-file /consul/client-token
```
pglass, line 34: Need to squash the output here so it doesn't print the token.

modules/mesh-task/templates/consul_client_command.tpl
```diff
@@ -26,10 +26,37 @@ login() {
...
+# This waits at most 2s (20 attempts with 0.1s sleep)
+COUNT=20
+while [ "$COUNT" -gt "0" ]; do
+    echo "Checking that the ACL token exists when reading it in the stale consistency mode ($COUNT attempts remaining)"
+    if read_token_stale; then
+        echo "Successfully read ACL token from the server"
+        break
+    fi
+    sleep 0.1
+    COUNT=$((COUNT - 1))
```
cthain, line 53: nit: can just use `((COUNT--))` here

cthain (approved):
Looks good Paul. I left a couple of nitpicks.

modules/mesh-task/templates/consul_client_command.tpl
pglass: Actually, the Consul image is Alpine-based, and doesn't include bash. It's just a posix shell (BusyBox shell, which is `ash`) and it doesn't seem to have the decrement operation.

PR #105, opened 2022-05-02
Placeholder for checking out tag in examples

lkysow:
When we do our next release we need to update these checkout commands so that
users will use a tagged version of our examples, not main.

### DIFF
--- a/examples/admin-partitions/README.md
+++ b/examples/admin-partitions/README.md
@@ -33,6 +33,7 @@ Clone this repository:
 
 ```console
 $ git clone https://github.com/hashicorp/terraform-aws-consul-ecs.git
+$ git checkout tags/<latest-version>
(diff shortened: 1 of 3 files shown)


examples/admin-partitions/README.md
```diff
@@ -33,6 +33,7 @@ Clone this repository:
 
 ```console
 $ git clone https://github.com/hashicorp/terraform-aws-consul-ecs.git
+$ git checkout tags/<latest-version>
```
pglass, line 36: I think this should just be, e.g. `git checkout v0.4.0`? Or does `tags/<version>` work for you?

lkysow: they both work for me

pglass: Oh, this does work for me. I tried it yesterday and must have typoed the tag name or something.

pglass (approved):

PR #110, opened 2022-05-27
Add mesh gateway WAN federation support

cthain:
- Update the dev-server module to support WAN federation via mesh-gateways.
- Update the gateway-task module to allow datacenter federation across the WAN.
- Update the mesh gateway example:
  - Create two separate VPCs.
  - Create two separate Consul datacenters.
  - WAN federate the Consul datacenters via mesh gateways.
  - Demonstrate that a client service running in `dc1` can successfully call a server service upstream running in `dc2` via the mesh gateways.
- Note: This PR does not include support for ACLs. That will be added in a subsequent PR.

## How I've tested this PR:
- Manually using the mesh-gateway example.

:eyes: 

## Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added


modules/gateway-task/variables.tf
```diff
@@ -0,0 +1,240 @@
...
+  default     = ""
+}
+
+variable "kind" {
+  description = "Gateway kind."
+  type        = string
+
+  validation {
+    error_message = "Gateway kind must be one of 'mesh-gateway', 'terminating-gateway', 'ingress-gateway'."
+    condition     = contains(["mesh-gateway", "terminating-gateway", "ingress-gateway"], var.kind)
```
erichaberkorn, line 202: What do you think about only allowing `mesh-gateway` for now?

modules/gateway-task/variables.tf
```diff
@@ -0,0 +1,240 @@
...
+  default     = []
+}
+
+variable "consul_service_meta" {
+  description = "A map of metadata that will be used for the Consul service registration"
+  type        = map(string)
+  default     = {}
+}
+
+variable "consul_namespace" {
```
erichaberkorn, line 24: I believe setting this to anything but `default` breaks things https://www.consul.io/docs/connect/config-entries/mesh#namespace

modules/gateway-task/validation.tf
```diff
@@ -0,0 +1 @@
+
```
erichaberkorn, line 1: Should we remove this file since it is empty?

examples/mesh-gateways/vpc.tf
```diff
@@ -0,0 +1,40 @@
...
+}
+
+data "aws_availability_zones" "available" {
+  filter {
+    name   = "opt-in-status"
+    values = ["opt-in-not-required"]
+  }
+}
+
+module "dc1_vpc" {
```
erichaberkorn, line 14: It isn't a big deal, but we could use [`count`](https://www.terraform.io/language/meta-arguments/count#basic-syntax) to reduce some of the duplication here.

examples/mesh-gateways/gateway/gateway.tf
```diff
@@ -0,0 +1,94 @@
...
+  port              = "8443"
+  protocol          = "TCP"
+
+  default_action {
+    type             = "forward"
+    target_group_arn = aws_lb_target_group.mesh_gateway.arn
+  }
+}
+
+// TODO: Ingress to each mesh gateway should only be allowed from the other
```
erichaberkorn, line 79: I think this is fine for an example

examples/mesh-gateways/datacenters.tf
```diff
@@ -0,0 +1,143 @@
...
+
+# resource "aws_security_group_rule" "ingress_from_dc2" {
+#   description              = "Access from dc2"
+#   type                     = "ingress"
+#   from_port                = 0
+#   to_port                  = 0
+#   protocol                 = "-1"
+#   source_security_group_id = module.dc2.dev_consul_server.security_group_id
+#   security_group_id        = module.dc1.dev_consul_server.security_group_id
+# }
```
erichaberkorn, line 120: Can this be removed?

examples/mesh-gateways/datacenter/consul-server.tf
```diff
@@ -0,0 +1,42 @@
+# Run the Consul dev server as an ECS task.
+module "dev_consul_server" {
+  name                        = "${var.name}-consul-server"
+  source                      = "../../../modules/dev-server"
+  datacenter                  = var.datacenter
+  primary_datacenter          = var.primary_datacenter
+  retry_join_wan              = var.retry_join_wan
+  primary_gateways            = var.primary_gateways
+  ecs_cluster_arn             = aws_ecs_cluster.this.arn
+  service_discovery_namespace = var.datacenter // TODO mesh gateway WAN fed requires the service discovery ns to be the datacenter? may need to remove this..
```
erichaberkorn, line 10: I think just setting this to `datacenter` is fine since the dev server is only for testing.

examples/mesh-gateways/datacenter/variables.tf
```diff
@@ -0,0 +1,90 @@
+variable "name" {
+  description = "Name to be used on all the resources as identifier."
+  type        = string
+}
+
+variable "datacenter" {
```
erichaberkorn, line 6: What do you think about making this `consul_datacenter` [to match mesh task](https://github.com/hashicorp/terraform-aws-consul-ecs/blob/main/modules/mesh-task/variables.tf#L298)?

modules/gateway-task/variables.tf
cthain: Yes, that is consistent with the change that we made to the `consul-ecs` config schema. It only allows `mesh-gateway`.

modules/gateway-task/validation.tf
cthain: I've added a validation to ensure that only one of `var.retry_join_wan` or `var.enable_mesh_gateway_wan_federation` is provided.

examples/mesh-gateways/gateway/gateway.tf
cthain: OK, I've added a note saying that this is an insecure setup for example only. In a production environment WAN traffic should be limited to only the other MGW.

modules/gateway-task/README.md
```diff
@@ -0,0 +1,7 @@
+# Gateway Task
+
+This module deploys a Consul gateway task as part of the service mesh.
+It is designed so you can easily modify existing [`ecs_task_definition`](https://registry.terraform.io/providers/hashicorp/aws/latest/docs/resources/ecs_task_definition)
+resources to use this module instead.
```
pglass, line 5: We should update this description. It could clarify which gateway types are currently supported.

modules/gateway-task/config.tf
```diff
@@ -0,0 +1,43 @@
...
+    consulCACertFile = var.consul_https_ca_cert_arn != "" ? "/consul/consul-https-ca-cert.pem" : ""
+    consulLogin      = local.consulLogin
+    service = {
+      name      = local.service_name
+      tags      = var.consul_service_tags
+      port      = 0 // note: field is required by consul-ecs schema. Not used for gateway.
+      meta      = local.consul_service_meta
+      namespace = var.consul_namespace
+      partition = var.consul_partition
+    }
```
pglass, line 25: Are we using this `service` block to configure these fields for the gateway service?

I was thinking we'd use only the `gateway` field to configure gateways. All of these `service.*` fields were duplicated on to the `gateway` object in the schema.json for that reason: https://github.com/hashicorp/consul-ecs/blob/4bf464a8df2645f1e28e1168ffa2f2bbcd2ef82a/config/schema.json#L349-L430

What do you think?

modules/gateway-task/variables.tf
```diff
@@ -0,0 +1,246 @@
...
+  default     = 256
+}
+
+variable "memory" {
+  description = "Amount (in MiB) of memory used by the task."
+  type        = number
+  default     = 512
+}
+
+variable "task_role" {
```
pglass, line 60: Could we remove the task_role and execution_role variables here? They seem unlikely to be needed for gateway tasks since there is no user app running in the task.

cthain: I think so.. It would simplify the `gateway-task` module if we alway use the generated task and execution roles.

examples/mesh-gateways/vpc.tf
cthain: I will include this in my next PR for enabling ACL support.

pglass (approved):

PR #116, opened 2022-06-09
Add ACL support for mesh gateway WAN federation

cthain:
- Update the `dev-server` module to support ACLs for mutl-dc WAN federation via mesh gateways.
- Update the `mesh-task` and `gateway-task` modules to allow passing a primary datacenter and replication token.
- Update the mesh-gateway example:
  - Add the `acl-controller` to both ECS clusters
  - Pass primary datacenter and replication token to client and server apps
  - Allow the `consul_ecs_image` to be passed to all tasks to support testing with dev images.

## How I've tested this PR:
- Unit tests
- Manually with the mesh-gateway example

:eyes: 

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added


modules/gateway-task/variables.tf
```diff
@@ -163,12 +163,24 @@ variable "acls" {
   default     = false
 }
 
+variable "enable_acl_token_replication" {
+  type        = bool
+  description = "Whether or not to enable ACL token replication for federated. ACL token replication is required when the gateway-task is part of a WAN federated Consul service mesh."
```
pglass, line 168:
```suggestion
  description = "Whether or not to enable ACL token replication. ACL token replication is required when the gateway-task is part of a WAN-federated Consul service mesh."
```

modules/gateway-task/variables.tf
```diff
@@ -163,12 +163,24 @@ variable "acls" {
...
+
 variable "consul_datacenter" {
   type        = string
   description = "The name of the Consul datacenter the client belongs to."
   default     = "dc1"
 }
 
+variable "consul_primary_datacenter" {
+  type        = string
+  description = "The name of the primary Consul datacenter. Required when the gateway-task is part of a WAN federated Consul service mesh."
```
pglass, line 180:
```suggestion
  description = "The name of the primary Consul datacenter. Required when the gateway-task is part of a WAN-federated Consul service mesh."
```

test/acceptance/tests/basic/terraform/mesh-gateway-validate/main.tf
```diff
@@ -0,0 +1,34 @@
...
+  region = "us-west-2"
+}
+
+variable "kind" {
+  type = string
+}
+
+variable "consul_namespace" {
+  type = string
+}
```
pglass, line 11: Should we remove the `consul_namespace` variable from gateway-task, since gateways always go in the default namespace?

pglass (approved):
Looks good!

pglass:
We need a changelog entry for the new gateway-task module as well

test/acceptance/tests/basic/terraform/mesh-gateway-validate/main.tf
cthain: Yes, I think this suggestion is a better approach than having the variable and validating that it is `""` or `"default"`. I will remove it.

CHANGELOG.md
```diff
@@ -1,3 +1,20 @@
+## Unreleased
+
+FEATURES
+* modules/gateway-task, modules/mesh-task, modules/dev-server, examples/mesh-gateways:
```
pglass, line 4: nit: We haven't included changes to the examples in the changelog before.

In general, I think we shouldn't include examples in the changelog. What do you think?

PR #119, opened 2022-06-15
Add load balancer and ECS service to gateway-task

cthain:
- Add an optional configuration to have the `gateway-task` module automatically create and configure a Network Load Balancer for public ingress.
- Update the `gateway-task` module to create the ECS service definition.

## How I've tested this PR:
- Unit tests
- Tested against the mesh-gateway example

:eyes: 

## Checklist:
- [x] Tests added
- [x] CHANGELOG entry added

### DIFF
--- a/modules/gateway-task/main.tf
+++ b/modules/gateway-task/main.tf
@@ -2,7 +2,7 @@ data "aws_region" "current" {}
 
 locals {
   // Must be updated for each release, and after each release to return to a "-dev" version.
-  version_string = "0.4.1-dev"
+  version_string = "0.5.0-dev"
 
   gossip_encryption_enabled = var.gossip_key_secret_arn != ""
   consul_data_volume_name   = "consul_data"
@@ -38,7 +38,16 @@ locals {
     }
   )
 
-  healthCheckPort = var.lan_port != 0 ? var.lan_port : 8443
+
+  lan_port    = var.lan_port != 0 ? var.lan_port : 8443
+  wan_port    = var.wan_port != 0 ? var.wan_port : local.lan_port
+  wan_address = var.lb_enabled ? aws_lb.this[0].dns_name : var.wan_address
+
+  load_balancer = var.lb_enabled ? [{
+    target_group_arn = aws_lb_target_group.this[0].arn
+    container_name   = "sidecar-proxy"
+    container_port   = local.lan_port
+  }] : []
 }
 
 resource "aws_ecs_task_definition" "this" {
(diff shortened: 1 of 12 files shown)


modules/gateway-task/main.tf
```diff
@@ -202,3 +211,70 @@ resource "aws_ecs_task_definition" "this" {
...
+  protocol          = "TCP"
+
+  default_action {
+    type             = "forward"
+    target_group_arn = aws_lb_target_group.this[0].arn
+  }
+}
+
+resource "aws_security_group_rule" "lb_ingress_rule" {
+  count             = var.lb_enabled ? length(var.security_groups) : 0
```
pglass, line 272: I'm thinking we shouldn't modify the security groups they pass in, unless they request it. (Also, we'd only need to add the rule on one of the security groups. AWS takes applies the union/sum of the rules from all the security groups)

Some thoughts:

1. We support creating a security group for them. We would add `lb_create_security_group = true` (default true). If true, we create a group + rule and append it to `var.security_groups`.
1. Also, add `lb_modify_security_group = true` and `lb_modify_security_group_id = <group-id>`. Then we would attach a rule to the specific security group they specify.
1. ???

What do you think?

cthain: I agree that we probably shouldn't just modify the security groups that are passed in. I really like the idea of doing 1.

Maybe we do 2. if there is enough time before the release?

pglass: Agreed, I think we should change this to do (1).

We can skip (2). And just document that they need to add a security group rule to allow ingress to the task from the load balancer.

pglass (approved):
Looks good!

The security group change can come in a follow up PR

modules/gateway-task/main.tf
cthain: I ended up doing both (1) and (2) because it was pretty straight forward.

modules/gateway-task/outputs.tf
```diff
@@ -22,3 +22,10 @@ output "task_tags" {
   value = aws_ecs_task_definition.this.tags_all
 }
```
pglass, line 24: We could maybe use an output for the security group id that is created for the LB, if it is created by this module?

modules/gateway-task/main.tf
```diff
@@ -70,135 +84,256 @@ resource "aws_ecs_task_definition" "this" {
...
+  from_port         = 0
+  to_port           = 0
+  protocol          = "-1"
+  cidr_blocks       = ["0.0.0.0/0"]
+  ipv6_cidr_blocks  = ["::/0"]
+  security_group_id = aws_security_group.this[0].id
+}
+
+resource "aws_security_group_rule" "lb_ingress_rule" {
+  count             = (var.lb_enabled && var.lb_create_security_group) || var.lb_modify_security_group ? 1 : 0
```
pglass, line 331: nit: So that if we set lb_enabled = false, then this security group rule would not get created in any case:

```suggestion
  count             = var.lb_enabled && (var.lb_create_security_group) || var.lb_modify_security_group) ? 1 : 0
```

pglass (approved):

PR #126, opened 2022-06-30
Skip HCP tests on PRs

pglass:
This updates CI and acceptance tests to skip the HCP tests on PRs.

The way this works is:
* Pass either `-var enable_hcp=true|false` to setup-terraform to control whether an HCP Consul cluster is created.
* HCP tests look for the `enable_hcp` output variable in setup-terraform, and skip themselves if there is no HCP cluster.
* In CI, we set `enable_hcp=true` for `main` and `release/*` branches. Otherwise, it is set to false.

For the moment, no enterprise tests will run on PRs, which I'll address in a follow up using the dev-server.

## How I've tested this PR:
Acceptance tests

👀 

## Checklist:
- [x] Tests added
- [ ] ~CHANGELOG entry added~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -126,8 +126,6 @@ jobs:
           name: terraform destroy
           working_directory: test/acceptance/setup-terraform
           command: |
-            set -x
-
             VARS="-var tags={\"build_url\":\"$CIRCLE_BUILD_URL\"}"
             VARS+=' -var launch_type=<<parameters.launch_type>>'
             case $CIRCLE_BRANCH in
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -92,12 +92,10 @@ jobs:
           name: terraform init & apply
           working_directory: test/acceptance/setup-terraform
           command: |
-            set -x
-
             terraform init
 
             VARS="-var tags={\"build_url\":\"$CIRCLE_BUILD_URL\"}"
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -92,9 +92,18 @@ jobs:
...
+
             terraform init
 
-            terraform apply -var tags="{\"build_url\": \"$CIRCLE_BUILD_URL\"}" -auto-approve -var launch_type=<<parameters.launch_type>>
+            VARS="-var tags={\"build_url\":\"$CIRCLE_BUILD_URL\"}"
+            VARS+=' -var launch_type=FARGATE'
+            case $CIRCLE_BRANCH in
+                main | release/*) VARS+=" -var enable_hcp=true";;
+                *) VARS+=" -var enable_hcp=false";;
+            esac
```
pglass, line 104: I don't love this but I didn't find a better way to alter the command based on the git branch.

cthain (commented):
This is awesome Paul! :heart_eyes:

.circleci/config.yml
```diff
@@ -119,7 +126,16 @@ jobs:
           name: terraform destroy
           working_directory: test/acceptance/setup-terraform
           command: |
-            terraform destroy -auto-approve -var launch_type=<<parameters.launch_type>>
+            set -x
```
pglass, line 129: Wait, I need to remove this here too

PR #129, opened 2022-06-30
Parallelize TestBasic

pglass:
This updates TestBasic to run the `secure:true` and `secure:false` cases in parallel.

Right now, this only saves about 5 mins (~15 mins vs 20mins). More importantly, it enables us to add additional tests for enterprise dev-server without extending the test runtime (hopefully).

## How I've tested this PR:
Acceptance tests

👀 

## Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/test/acceptance/tests/basic/basic_test.go
+++ b/test/acceptance/tests/basic/basic_test.go
@@ -737,37 +737,64 @@ func TestValidation_MeshGateway(t *testing.T) {
 }
 
 func TestBasic(t *testing.T) {
-	cases := []bool{true, false}
-	for _, secure := range cases {
-		t.Run(fmt.Sprintf("secure: %t", secure), func(t *testing.T) {
+	cfg := suite.Config()
+
+	cases := []struct {
+		secure        bool
+		stateFile     string
+		ecsClusterARN string
+		datacenter    string
+	}{
+		// Works for 2 clusters in parallel. Each test case should have a unique cluster
+		// and requires a unique Terrform state file. The datacenter must be unique within
+		// the VPC to avoid conflicts in CloudMap namespaces.
+		{false, "terraform.tfstate", cfg.ECSClusterARN, "dc1"},
+		{true, "terraform-secure.tfstate", cfg.ECSCluster2ARN, "dc2"},
+	}
+	for _, c := range cases {
+		c := c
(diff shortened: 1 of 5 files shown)


test/acceptance/tests/basic/basic_test.go
```diff
@@ -737,37 +749,66 @@ func TestValidation_MeshGateway(t *testing.T) {
...
+		secure        bool
+		stateFile     string
+		ecsClusterARN string
+		datacenter    string
+	}{
+		// Works for 2 clusters in parallel. Each test case should have a unique cluster
+		// and requires a unique Terrform state file. The datacenter must be unique within
+		// the VPC to avoid conflicts in CloudMap namespaces.
+		{false, "terraform.tfstate", cfg.ECSClusterARN, "dc1"},
+		{true, "terraform-secure.tfstate", cfg.ECSCluster2ARN, "dc2"},
```
pglass, line 766: I didn't think too hard about this and just tried to made it work for the two test cases we have with the few code changes.

These tests share a VPC, so we have to worry about the CloudMap namespace being unique ("dc1", "dc2", etc). I opted to just explicitly specify the state file names and dc names and clusters here, since we'll have few of these test cases.

When we add test cases, we need to modify setup-terraform to create additional ECS clusters, which is fine except that for the EC2 launch type we'll need many instances (four t3a.micro instances per cluster).

I did consider:

* Having each test spin up its required resources. But this makes it a little worse to re-run tests locally, since more setup time is spent in the test case versus once in setup-terraform.
* Splitting test cases into more CI jobs. But this isn't useful for running tests in parallel locally, though.

Instead of specifying state files, we could:
* Use Terratest's [CopyTerraformFolderToDest](https://github.com/gruntwork-io/terratest/blob/583586f7fd0e0ec22a3d46e953df06e408d589dd/modules/test-structure/test_structure.go#L108) function. But I don't like the idea of my TF state ending up in some temp dir, in case I need to manually clean up things locally.
* Or, Terraform supports workspaces that isolate state, too, but I don't know how they work and didn't look into it for now. There's a Terratest PR open to support specifying workspaces (`https://github.com/gruntwork-io/terratest/pull/1102`) so this could be something we potentially switch to.

PR #131, opened 2022-07-13
Fixes to HCP test

sanon-dev:
- Fixes to HCP to use ClusterARN slices within validation test. 

## How I've tested this PR:
- Ran acceptance test locally

👀 

## Checklist:
- [x] Tests added
- [x] ~CHANGELOG entry added~ 

### DIFF
--- a/test/acceptance/tests/hcp/hcp_test.go
+++ b/test/acceptance/tests/hcp/hcp_test.go
@@ -615,7 +615,7 @@ func TestAuditLogging(t *testing.T) {
 		Namespace:    "default",
 		ConsulClient: consulClient,
 		Region:       cfg.Region,
-		ClusterARN:   cfg.ECSClusterARN,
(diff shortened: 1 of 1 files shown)


test/acceptance/tests/hcp/hcp_test.go
```diff
@@ -615,7 +615,7 @@ func TestAuditLogging(t *testing.T) {
 		Namespace:    "default",
```
pglass, line 615: nit: On line 600 (where I can't comment), you can also remove `"ecs_cluster_1_arn", "ecs_cluster_2_arn"` from the `ignoreVars`. These numbered cluster arns are no longer used by any of the Terraform modules, so we don't need to reference them (but having them in ignoreVars won't cause any issues either).

pglass (approved):
👍

cthain (approved):

PR #133, opened 2022-07-22
Fix gossip key ARN

cthain:
Fix a bug in the `dev-server` where it doesn't pick up the correct secret ARN for the gossip encryption key when the module is configured to generate the gossip key. When `var.gossip_encryption_enabled = true` the module is supposed to create and use a gossip key. However, when `gossip_encryption_enabled = true` it was always using the value of `var.gossip_key_secret_arn`, even if that was unset (which it should be in the case where the key is generated). This caused the creation of the execution role policy to fail because the `Resource` ARN for gossip key secret was the empty string.

## How I've tested this PR:
- :eyes:
- Using the Lambda registrator acceptance tests (how I found the issue)

:eyes:

## Checklist:
- [x] ~Tests added~ N/A
- [x] CHANGELOG entry added 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,13 @@ FEATURES
   variable for configuring login to the AWS IAM auth method.
   [[GH-132](https://github.com/hashicorp/terraform-aws-consul-ecs/pull/132)]
 
+BUG FIXES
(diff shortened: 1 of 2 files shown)


sanon-dev (approved):
Looks good 👍

pglass (approved):

PR #139, opened 2022-11-08
add fargate ha consul cluster backed by efs

fdr2:
- add an additional module implementing consul on ecs backed by efs storage
- add a load test for consul kv

## How I've tested this PR:
After building consul with [PR: 13782](https://github.com/hashicorp/consul/pull/13782) to support ECS for Go Discover, create a variables file as described within the [README.md](https://github.com/fdr2/terraform-aws-consul-ecs/blob/5ed316420879f9731ef5a4c68812643e19cd92b5/examples/ha-cluster-fargate/README.md) to point your consul image variable at the image url for the image you have built (until consul is updated with the necessary go-discover version).

Using the example `ha-cluster-fargate`, create ca and certs, deploy resources to AWS and load test.

## Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 
- [X] Tag @pglass for visibility

### DIFF
--- a/modules/ecs-service/variables.tf
+++ b/modules/ecs-service/variables.tf
@@ -1,0 +1,125 @@
+variable "name" { default = "container" }
+variable "tags" {
+  description = "Tags for the resources"
+  type        = map(string)
+  default = {
+    env : "dev"
+  }
+}
+
+variable "ecs_cluster_name" {
+  description = "ECS Cluster Name to Deploy this container and launch template"
+  type        = string
+}
+
+variable "ecs_execution_role_arn" {
+  type = string
+}
+
+variable "ecs_task_role_arn" {
+  type = string
+}
+
+variable "ecs_task_role_id" {
+  type = string
+}
+
+variable "vpc_id" {
+  description = "VPC ID the subnets are attached to"
+  type        = string
+}
+
+variable "subnet_ids" {
+  description = "List of subnet ids to deploy the containers in"
+  type        = list(string)
+}
+
+variable "desired_count" {
+  description = "ECS Service Fixed Count"
+  type        = number
+  default     = 1
+}
+
+variable "cpu" {
+  default = 256
+}
+
+variable "memory" {
+  default = 512
+}
+
+variable "deployment_maximum_percent" {
+  default = 200
+}
+variable "deployment_minimum_healthy_percent" {
+  default = 100
+}
+variable "deployment_circuit_breaker_enable" {
+  default = false
+}
+
+variable "security_group_ids" {
+  description = "Additional Security Group IDs to add to the Service"
+  type        = list(string)
+  default     = []
+}
+
+variable "secrets" {
+  type    = list(object({ name : string, valueFrom : string }))
+  default = []
+}
+
+variable "task_definition" {
+  default = {}
+}
+
+variable "operating_system_family" {
+  default = "LINUX"
+}
+
+variable "cpu_architecture" {
+  default = "X86_64"
+}
+
+variable "efs_volumes" {
+  type = map(object({
+    file_system_id : string
+    access_point_id : string
+    root_directory : string
+    transit_encryption : string
+    encryption_port : number
+    iam : string
+  }))
+  default = {}
+}
+
+variable "container_name" {
+  type    = string
+  default = ""
+}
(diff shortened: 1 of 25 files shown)


modules/ha-cluster/variables.tf
```diff
@@ -0,0 +1,328 @@
...
+variable "run_k6" {
+  type    = bool
+  default = false
+}
+
+variable "k6_apikey" {
+  description = "K6 API key"
+  type        = string
+  default     = ""
+}
```
pglass, line 150: Please remove these `k6` variables (I don't see that they are used anywhere in the `ha-cluster` module).

modules/ha-cluster/main.tf
```diff
@@ -0,0 +1,894 @@
...
+    protocol : "tcp"
+    }, {
+    containerPort : 8501,
+    hostPort : 8501,
+    protocol : "tcp"
+    }, {
+    containerPort : 8502,
+    hostPort : 8502,
+    protocol : "tcp"
+    }, {
```
pglass, line 196: I notice you have `ports.grpc_tls = 8503` above, so probably add 8503 to the list of container ports here for consistency. (Note that `ports.grpc_tls` is only supported in Consul 1.14+)

modules/ha-cluster/main.tf
```diff
@@ -0,0 +1,894 @@
...
+  %{for dnsname in var.additional_dns_names~}
+    -additional-dnsname="${dnsname}" \
+  %{endfor~}
+%{endif~}
+EOF
+
+  consul_portmap = [{
+    containerPort : 8300,
+    hostPort : 8300,
+    protocol : "tcp"
```
pglass, line 175: nit: Please remove the `hostPort` field from these port mappings. We only support the `awsvpc` network mode for our other modules, so we can assume the same here (for now). In `awsvpc` network mode, the [host port cannot be different than the container port](https://docs.aws.amazon.com/AmazonECS/latest/APIReference/API_PortMapping.html), so I'd prefer to leave the field out (similar to the [dev-server module](https://github.com/hashicorp/terraform-aws-consul-ecs/blob/a8cdec6b6ce781679033822acab23f5b58a2129a/modules/dev-server/main.tf#L163-L173))

modules/ecs-service/variables.tf
```diff
@@ -0,0 +1,125 @@
...
+  }))
+  default = {}
+}
+
+variable "container_name" {
+  type    = string
+  default = ""
+}
+
+variable "sidecar_name" {
```
pglass, line 101: The `sidecar_name` variable appears to be unused? Can it be removed?

modules/ha-cluster/variables.tf
```diff
@@ -0,0 +1,328 @@
...
+
+variable "datadog_task_cpu" {
+  description = "Set the Consul Server task CPU limit. Default is 1792."
+  default     = 256
+}
+
+variable "datadog_task_memory" {
+  description = "Set the Consul Server container Memory limit. Default is 3584."
+  default     = 512
+}
```
pglass, line 129: Please remove the datadog specific functionality.

Instead, we can support generic variables, like `additional_container_definitions`, `additional_task_role_policies`, etc, in order to inject an additional container into the task definition. That way we can support any kind of additional "sidecar" container.

modules/ha-cluster/variables.tf
```diff
@@ -0,0 +1,328 @@
...
+
+variable "lb_ingress_rule_security_groups" {
+  description = "Security groups that are allowed access to the load balancer."
+  type        = list(string)
+  default     = null
+}
+
+variable "consul_count" {
+  description = "Number of consul servers to deploy. Default 3"
+  default     = 3
```
pglass, line 172: minor / non-blocking: Validate the consul_count is odd and greater than 0. Something like the following:

```suggestion
  default     = 3
  
  validation {
    error_message = "The consul_count must be odd and must be greater than 0."
    condition     = var.consul_count % 2 == 1 && var.consul_count > 0
  }
```

pglass (commented):
Hi @fdr2. I've (finally) taken an initial pass through this, but have not reviewed everything. 

Thank you so much for doing this!

PR #141, opened 2022-11-15
Update test dependencies

pglass:

- Bump acceptance tests to Go 1.19
- Bump Terraform to 1.3.4
- Update other versions (gotestsum, golangci-lint, aws cli, session-manager-plugin)
- Don't use `grpc_tls` setting yet, which is not valid in 1.13.3

## How I've tested this PR:

* Acceptance test locally with both Consul 1.12.2 and 1.13.3

## Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,7 +51,7 @@ jobs:
           name: lint-consul-retry
           working_directory: test/acceptance
           command: |
-            go get -u github.com/hashicorp/lint-consul-retry
+            go install github.com/hashicorp/lint-consul-retry@latest
             lint-consul-retry
 
       - run:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ orbs:
 executors:
   consul-ecs-test:
     docker:
-      - image: docker.mirror.hashicorp.services/hashicorpdev/consul-ecs-test:0.3.2
+      - image: docker.mirror.hashicorp.services/hashicorpdev/consul-ecs-test:0.3.3
     environment:
       TEST_RESULTS: &TEST_RESULTS /tmp/test-results # path to where test results are saved
     # medium (2cpu / 4gb) with parallel invocations of Terraform sometimes ran out of memory.
@@ -48,11 +48,6 @@ jobs:
             fi
 
       - run:
-          name: go vet
-          working_directory: test/acceptance
-          command: go vet ./...
-
-      - run:
           name: lint-consul-retry
           working_directory: test/acceptance
           command: |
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -47,24 +47,17 @@ jobs:
               exit 1
             fi
 
-      - run:
-          name: go vet
-          working_directory: test/acceptance
-          command: go vet ./...
-
```
pglass, line 54: removed, because golangci-lint runs vet

.circleci/config.yml
```diff
@@ -47,24 +47,17 @@ jobs:
...
           command: |
-            go get -u github.com/hashicorp/lint-consul-retry
+            go install github.com/hashicorp/lint-consul-retry@latest
             lint-consul-retry
 
       - run:
           name: golangci-lint
           working_directory: test/acceptance
-          command: |
-            curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.31.0
```
pglass, line 66: Removed golangci-lint install here, because it is already included in the consul-ecs-test:0.3.3 image

modules/dev-server/main.tf
```diff
@@ -401,7 +401,7 @@ exec consul agent -server \
   -hcl='cert_file = "/consul/${var.datacenter}-server-consul-0.pem"' \
   -hcl='key_file = "/consul/${var.datacenter}-server-consul-0-key.pem"' \
   -hcl='auto_encrypt = {allow_tls = true}' \
-  -hcl='ports { https = 8501, grpc_tls = 8502 }' \
```
pglass, line 404: This is here because I merged https://github.com/hashicorp/terraform-aws-consul-ecs/pull/138. But `grpc_tls` is not valid in 1.13.3

test/acceptance/go.mod
```diff
@@ -11,3 +11,46 @@ require (
 	github.com/hashicorp/consul/sdk v0.9.0
 	github.com/stretchr/testify v1.4.0
 )
+
+require (
```
cthain, line 15: I guess this is just due to the version bump to go 1.19

cthain (approved):

test/acceptance/go.mod
pglass: Yeah, it looks like [in 1.17](https://go.dev/doc/go1.17#go-command) they started to include indirect dependencies in go.mod:

> Because the number of explicit requirements may be substantially larger in an expanded Go 1.17 go.mod file, the newly-added requirements on indirect dependencies in a go 1.17 module are maintained in a separate require block from the block containing direct dependencies.

PR #146, opened 2022-12-07
Test with Consul 1.13 and 1.14

pglass:

Update the acceptance tests and CI jobs to test with Consul 1.13 and 1.14.

## How I've tested this PR:

- [x] CI tests
- [ ] HCP acceptance tests locally

## Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -70,8 +70,12 @@ jobs:
     executor: consul-ecs-test
     parameters:
       launch_type:
-        default: ""
         type: string
+      consul_version:
+        type: string
+      # HCP is never enabled for PRs.
+      enable_hcp:
+        type: boolean
     steps:
       - checkout
 
@@ -94,9 +98,9 @@ jobs:
             terraform init
 
             VARS="-var tags={\"build_url\":\"$CIRCLE_BUILD_URL\"}"
-            VARS+=' -var launch_type=<<parameters.config.launch_type>>'
+            VARS+=' -var launch_type=<<parameters.launch_type>>'
             case $CIRCLE_BRANCH in
-                main | release/*) VARS+=" -var enable_hcp=<<parameters.config.enable_hcp>>";;
+                main | release/*) VARS+=" -var enable_hcp=<<parameters.enable_hcp>>";;
                 *) VARS+=" -var enable_hcp=false";;
             esac
 
@@ -126,14 +130,23 @@ jobs:
           working_directory: test/acceptance/setup-terraform
           command: |
             VARS="-var tags={\"build_url\":\"$CIRCLE_BUILD_URL\"}"
-            VARS+=' -var launch_type=<<parameters.config.launch_type>>'
+            VARS+=' -var launch_type=<<parameters.launch_type>>'
             case $CIRCLE_BRANCH in
-                main | release/*) VARS+=" -var enable_hcp=<<parameters.config.enable_hcp>>";;
+                main | release/*) VARS+=" -var enable_hcp=<<parameters.enable_hcp>>";;
                 *) VARS+=" -var enable_hcp=false";;
             esac
 
             terraform destroy -auto-approve $VARS
           when: always
+
+
+acceptance-common: &acceptance-common
+  filters:
+    branches:
+      # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
+      ignore: /pull\/[0-9]+|docs\/.*/
+  requires:
+    - go-fmt-and-lint-acceptance
 
 workflows:
   version: 2.1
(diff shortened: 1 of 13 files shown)


.circleci/config.yml
```diff
@@ -127,29 +132,37 @@ jobs:
...
-            - go-fmt-and-lint-acceptance
+      # We have a limit of 6 HCP Consul clusters.
+      # The following controls whether to enable HCP when testing release branches.
+      # HCP is always disabled for tests on PRs.
+      - acceptance: {name: "acceptance-1.12-FARGATE-HCP", consul_version: '1.12.7', enable_hcp: true, launch_type: FARGATE, <<: *acceptance-common}
+      - acceptance: {name: "acceptance-1.13-FARGATE", consul_version: '1.13.4', enable_hcp: false, launch_type: FARGATE, <<: *acceptance-common}
+      - acceptance: {name: "acceptance-1.14-FARGATE-HCP", consul_version: '1.14.1', enable_hcp: true, launch_type: FARGATE, <<: *acceptance-common}
+      - acceptance: {name: "acceptance-1.12-EC2", consul_version: '1.12.7', enable_hcp: false, launch_type: EC2, <<: *acceptance-common}
+      - acceptance: {name: "acceptance-1.13-EC2-HCP", consul_version: '1.13.4', enable_hcp: true, launch_type: EC2, <<: *acceptance-common}
+      - acceptance: {name: "acceptance-1.14-EC2", consul_version: '1.14.1', enable_hcp: false, launch_type: EC2, <<: *acceptance-common}
```
pglass, line 167: Here, I didn't use the full matrix of `launch_type x consul_version` because that would require all 6 of our HCP clusters, per release branch.

I settled for 3 HCP clusters at a time, so this will test all three Consul versions. Two of those HCP enabled test runs use Fargate, and one uses EC2. (Note that we always test with the dev server in each job, even when HCP is enabled)

For the config, the alternative is to use [matrix](https://circleci.com/docs/configuration-reference/#matrix) + excludes. But I felt that being explicit about what we _are_ testing is a little clearer (but uglier) versus figuring out all the configurations we need to exclude. Also, we don't need to run that same config twice with both HCP enabled and disabled (because we always run the dev server tests regardless).

.circleci/config.yml
```diff
@@ -127,29 +132,37 @@ jobs:
...
-          filters:
-            branches:
-              # Forked pull requests have CIRCLE_BRANCH set to pull/XXX
-              ignore: /pull\/[0-9]+|docs\/.*/
-          requires:
-            - go-fmt-and-lint-acceptance
+      # We have a limit of 6 HCP Consul clusters.
+      # The following controls whether to enable HCP when testing release branches.
+      # HCP is always disabled for tests on PRs.
+      - acceptance: {name: "acceptance-1.12-FARGATE-HCP", consul_version: '1.12.6', enable_hcp: true, launch_type: FARGATE, <<: *acceptance-common}
```
pglass, line 162: One thing that is confusing: The job will be named `acceptance-1.12-FARGATE-HCP`. When the job runs on release branches, it will test with HCP, but we never enable HCP for tests on PRs. 

I'm sure when someone else looks at the job name, they will think that it is running tests with HCP on their PR, which isn't the case.

modules/dev-server/main.tf
```diff
@@ -401,7 +401,11 @@ exec consul agent -server \
   -hcl='cert_file = "/consul/${var.datacenter}-server-consul-0.pem"' \
   -hcl='key_file = "/consul/${var.datacenter}-server-consul-0-key.pem"' \
   -hcl='auto_encrypt = {allow_tls = true}' \
+%{if local.is_consul_1_14_plus~}
+  -hcl='ports { https = 8501, grpc_tls = 8503 }' \
+%{else~}
   -hcl='ports { https = 8501, grpc = 8502 }' \
+%{endif~}
```
pglass, line 408: A note on this: I didn't need any changes to the Consul client agent config, because we do not enable TLS locally within the task. The client agent is configured with [`ports.grpc = 8502`](https://github.com/hashicorp/terraform-aws-consul-ecs/blob/8986f2c418cd389638fc9ea44f0a1857f61b5988/modules/mesh-task/templates/consul_agent_defaults.hcl.tpl#L16) and [mesh-init](https://github.com/hashicorp/consul-ecs/blob/76f957ed59f7fa0fcf48f000628b0cbc2957c256/subcommand/mesh-init/command.go#L127) configures Envoy with that port (8502).

.circleci/config.yml
pglass: Not sure if it's coincidence, but the `acceptance-1.14-EC2` job has flaked several times. A couple times exhibited a CloudWatch parsing bug, which I've fixed - but this might indicate some underlying issue. Another time it failed seemingly without a specific error that I could find. Another time it showed an execute-command session timeout, but I was unsure of the cause.

But when I retried the 1.14 EC2 job, it passed each time. Also, the 1.14 Fargate job never failed. I've also run that test several times locally, and never saw a failure. 🤔 

I have an inkling there might be some kind of AWS limit we are hitting, either in session manager or in cloud watch. Maybe more parallel things happening (6 jobs * 3 parallel test cases) when all 6 jobs are running, compared to when I re-run a single job or locally in a test run.

.circleci/config.yml
cthain: Yes, I noticed this in the pipeline but I think it is OK because we just don't test HCP on PRs period. I like the approach that you've taken here to be explicit about what the tests are doing and I think it is a good tradeoff.

.circleci/config.yml
cthain: I saw the flakes (I even restarted one of the jobs). It does **seem** to just be flakes because it is not consistently failing for any specific reason.

PR #164, opened 2023-03-22
Migrate from CircleCI to Github Actions

cthain:
- Remove CircleCI config
- Add Github actions config
- Add codeowners file
- Add dependabot for GHA
- Fix some acceptance tests with race conditions
- Remove custom docker test image because it's no longer required.

## How I've tested this PR:
- Acceptance tests running in GHA

:eyes: 

## Checklist:
- [x] Tests updated
- [x] ~CHANGELOG entry added~ Not applicable


kisunji (approved):
Not super familiar with .github but looks good! Should we re-run `copywrite` on your new CI files? I'm sure you can manually copy-paste as well.

.github/workflows/terraform-ci.yml
```diff
@@ -0,0 +1,218 @@
...
+  CONSUL_LICENSE: ${{ secrets.CONSUL_LICENSE }}
+  HCP_CLIENT_ID: ${{ secrets.HCP_CLIENT_ID }}
+  HCP_CLIENT_SECRET: ${{ secrets.HCP_CLIENT_SECRET }}
+permissions: {}
+jobs:
+  action-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@ac593985615ec2ede58e132d2e21d2b1cbd6127c # v3.3.0
```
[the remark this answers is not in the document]
cthain: Ya, this is a security requirement. I assume it is because strictly speaking tags are mutable whereas commits are immutable.

.github/workflows/terraform-ci.yml
```diff
@@ -0,0 +1,218 @@
...
+      - name: Setup Go
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
+        with:
+          go-version: ${{ needs.get-go-version.outputs.go-version }}
+      - name: go fmt
+        run: |
+          files=$(go fmt ./...)
```
pglass, line 62: minor: golangci-lint does (or can do) a gofmt check for us, so we could potentially remove this step.

cthain:
> Should we re-run `copywrite` on your new CI files? I'm sure you can manually copy-paste as well.

Good catch :facepalm: .. I forgot about the copywrite headers. I'll add them to the new files manually :+1:

pglass (approved):

.github/workflows/terraform-ci.yml
```diff
@@ -0,0 +1,218 @@
...
+      if: always()
+      run: |
+        cd setup-terraform/
+        VARS="-var tags={\"build_url\":\"$GITHUB_SERVER_URL/$GITHUB_REPOSITORY/actions/runs/$GITHUB_RUN_ID\"}"
+        VARS+=' -var launch_type=${{ matrix.launch-type }}'
+        VARS+=' -var consul_version=${{ matrix.consul-version }}'
+        case $GITHUB_REF_NAME in
+            main | release/*) VARS+=" -var enable_hcp=${{ matrix.enable-hcp }}";;
+            *) VARS+=" -var enable_hcp=false";;
+        esac
```
pglass, line 217: There might be a cleaner way to accomplish this in Github actions than there was in Circle CI.

Maybe:

* Create a workflow the runs for release branches and main
* Create another workflow for pull requests / other branches

Both of those workflows can trigger a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) that passes in an `enable_hcp=true/false` setting and runs acceptance tests.

(Probably not worth the effort)

cthain: This is a good idea and I did give it some thought. I think we should punt this down the road for now because I'm supposed to have this all wrapped up by Friday (Mar 24th.)

.github/workflows/terraform-ci.yml
cthain: Great idea! The `gofmt` linter isn't enabled by default but it can be by passing the `--enable gofmt` args. I'll add this and remove the expliclt `go fmt` step.

PR #186, opened 2023-07-20
Fix acceptance tests

Ganeshrockz:
- Bump the vpc module to `5.0.0` because the existing version was failing with some unknown input variable related error.
- Bump consul to the latest patch versions for each minor version release.
- For some reason, the default vpc did not contain security group rules for egress and ingress traffic. Added rules to cover the same

## Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added (Not needed)


cthain (approved):
Looks good :+1:

PR #188, opened 2023-07-24
Modify mesh-task and controller

Ganeshrockz:
- Makes changes for mesh-task and controller modules to align with the consul-dataplane based architecture in ECS.
- Changed the named of `acl-controller` module to just `controller`
- Removed client agent specific TF resources and tests.
- Gateway related flows will be changed in the upcoming PRs.
- HCP related changes will be made in the upcoming PRs. CI will also not test HCP.

## How I've tested this PR:

Unit and integration tests.

## Checklist:
- [x] Tests added
- [ ] CHANGELOG entry added (Not needed. Will push this to a non default branch)


modules/mesh-task/variables.tf
```diff
@@ -477,8 +441,36 @@ variable "consul_ecs_config" {
...
+  validation {
+    error_message = "Only the 'port', 'https', 'tls', 'tlsServerName' and 'caCertFile' fields are allowed in http_tls_config."
+    condition = alltrue([
+      for key in keys(var.http_tls_config) :
+      contains(["port", "https", "tls", "tlsServerName", "caCertFile"], key)
+    ])
+  }
 }
+
+variable "grpc_tls_config" {
```
Ganeshrockz, line 461: Exposing first class variables for every field in gRPC and http tls settings seem like an overkill. Exposing them as first class config variables which we will merge with the default setting in `config.tf`. Open for feedback on this.

modules/mesh-task/variables.tf
```diff
@@ -317,53 +284,50 @@ variable "tags" {
...
 variable "tls" {
-  description = "Whether to enable TLS for the mesh-task for the control plane traffic."
+  description = "Whether to enable TLS for the communication between mesh-task and Consul's HTTP and gRPC interfaces."
   type        = bool
   default     = false
 }
 
-variable "consul_server_ca_cert_arn" {
-  description = "The ARN of the Secrets Manager secret containing the Consul server CA certificate for Consul's internal RPC."
+variable "tls_server_name" {
```
Ganeshrockz, line 292: Only the default TLS settings get first class TF input variables

modules/mesh-task/validation.tf
```diff
@@ -3,11 +3,9 @@
 
 locals {
   require_port_unless_no_port_true                          = !var.outbound_only && var.port == 0 ? file("ERROR: port must be set if outbound_only is false") : null
-  require_ca_cert_if_tls_enabled                            = (var.tls && var.consul_server_ca_cert_arn == "") ? file("ERROR: consul_server_ca_cert_arn must be set if tls is true") : null
```
Ganeshrockz, line 6: Removed CA cert related validation because we now have multiple ways to configure ca certs (Via Env vars, via files or protocol specific TLS settings which makes validation a lot more complex)

cthain (commented):
Left my initial comments from a quick pass over the PR. I will review in more detail and review the tests tomorrow.

modules/mesh-task/variables.tf
```diff
@@ -317,53 +284,50 @@ variable "tags" {
...
-variable "consul_datacenter" {
+variable "consul_server_ca_cert_arn" {
+  description = "The ARN of the Secrets Manager secret containing the Consul server CA certificate for Consul's internal RPC and HTTP interfaces."
   type        = string
-  description = "The name of the Consul datacenter the client belongs to."
-  default     = "dc1"
+  default     = ""
 }
 
-variable "consul_primary_datacenter" {
```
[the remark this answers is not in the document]
Ganeshrockz: We still support WAN federation. The primary datacenter gets passed to the agent configuration and not the gateway task itself. So the field becomes irrelevant in the new architecture

Ganeshrockz:
@cthain / @pglass Can you re-review this PR whenever possible? Thanks 🙇

modules/controller/variables.tf
```diff
@@ -0,0 +1,171 @@
...
+  default     = "/consul-ecs/"
+}
+
+variable "subnets" {
+  description = "Subnets where the controller task should be deployed. If these are private subnets then there must be a NAT gateway for image pulls to work. If these are public subnets then you must also set assign_public_ip for image pulls to work."
+  type        = list(string)
+}
+
+variable "consul_server_address" {
+  description = "Address of the consul server host"
```
pglass, line 55: This address can also be an `exec=` string. We should update this description to be similar to [consul-k8s](https://developer.hashicorp.com/consul/docs/k8s/helm#v-externalservers-hosts)

```suggestion
  description = "Address of the Consul servers. This can be an IP address, a DNS name, or an `exec=` string specifying a script that outputs IP addresses. Refer to https://github.com/hashicorp/go-netaddrs#summary for details. This variable should not specify the port. Instead, use var.http_config.port and var.grpc_config.port to change the server HTTP and gRPC ports.
```

pglass: nit: Consul Dataplane calls this setting `-addresses`, and consul-server-connection-manager also uses `Addresses` for the config field. K8s uses `externalServers.hosts` for the same setting (`hosts` was the field name used prior to Consul Dataplane)

IMO, using the plural, "addresses", is slightly confusing because it is not actually a list. To be consistent, should we make this plural as well - `consul_server_addresses`?

modules/mesh-task/main.tf
```diff
@@ -5,10 +5,9 @@ data "aws_region" "current" {}
 
 locals {
   // Must be updated for each release, and after each release to return to a "-dev" version.
-  version_string = "0.6.0"
+  version_string = "1.0.0-dev"
```
pglass, line 8: Are we sure about bumping this to 1.0.0 now?

modules/mesh-task/variables.tf
```diff
@@ -165,7 +171,7 @@ variable "envoy_public_listener_port" {
   }
 
   validation {
-    error_message = "The envoy_public_listener_port must not conflict with the following ports that are reserved for Consul and Envoy: 8300, 8301, 8302, 8500, 8501, 8502, 8600, 19000."
+    error_message = "The envoy_public_listener_port must not conflict with the following ports that are reserved for Consul and Envoy: 8300, 8301, 8302, 8500, 8501, 8502, 8600, 10000, 19000."
```
pglass, line 174: We can remove the validation for most of the Consul client agent ports here, I think: 8300, 8301, 8302, 8500, 8501, 8502. Without client agents

We should keep the DNS port reserved though (8600) because Consul Dataplane supports running a DNS server/proxy.

modules/mesh-task/variables.tf
pglass, line 1: Noting that the new variables here should generally match any changes made to the same variables in the controller module (I won't copy all my comments over).

test/acceptance/tests/basic/basic_test.go
```diff
@@ -19,27 +19,12 @@ import (
...
-		TerraformDir: "./terraform/ca-cert-validate",
-		NoColor:      true,
-	})
-	t.Cleanup(func() {
-		_, _ = terraform.DestroyE(t, terraformOptions)
-	})
-	_, err := terraform.InitAndPlanE(t, terraformOptions)
-	require.Error(t, err)
-	require.Contains(t, err.Error(), "ERROR: consul_server_ca_cert_arn must be set if tls is true")
-}
```
pglass, line 41: non-blocking / tech debt: This is long overdue, but it would be good to move the TestValidation cases to a separate file or separate test package from the TestBasic acceptance test.

test/acceptance/tests/basic/basic_test.go
```diff
@@ -19,27 +19,12 @@ import (
 	"github.com/gruntwork-io/terratest/modules/random"
 	"github.com/gruntwork-io/terratest/modules/shell"
 	"github.com/gruntwork-io/terratest/modules/terraform"
-	"github.com/hashicorp/consul/sdk/testutil/retry"
+	"github.com/hashicorp/serf/testutil/retry"
```
pglass, line 22: Make sure to use testutils from the consul repo (not serf).

```suggestion
	"github.com/hashicorp/consul/sdk/testutil/retry"
```

My editor magic also likes to switch this out for whatever reason.

test/acceptance/tests/basic/basic_test.go
```diff
@@ -401,50 +386,59 @@ func TestValidation_EnvoyPublicListenerPort(t *testing.T) {
...
 
 	terraformOptions := &terraform.Options{
-		TerraformDir: "./terraform/checks-validate",
+		TerraformDir: "./terraform/envoy-readiness-port-validate",
 		NoColor:      true,
 	}
 	terraform.Init(t, terraformOptions)
 
 	for name, c := range cases {
+		c := c
```
pglass, line 420: (One day soon, we won't need this loop capture hack anymore: https://github.com/golang/go/wiki/LoopvarExperiment)

test/acceptance/tests/basic/basic_test.go
```diff
@@ -664,148 +772,150 @@ func TestValidation_RolePath(t *testing.T) {
...
-				require.Contains(t, err.Error(), c.expError)
-			} else {
-				require.NoError(t, err)
-			}
-		})
-	}
-}
+// TODO: Revisit this test
+//
+// func TestValidation_MeshGateway(t *testing.T) {
```
pglass, line 777: nit / suggestion: You can add a `t.Skip` instead of commenting out the test if you prefer: `t.Skip("TODO(ganeshrocks): revisit this test for agentless")`. Then you'll see the skipped test in the test run output.

test/acceptance/tests/basic/basic_test.go
```diff
@@ -925,6 +1035,44 @@ func TestBasic(t *testing.T) {
...
 			})
 
+			var controllerTaskID string
+			if c.secure {
+				retry.RunWith(&retry.Timer{Timeout: 2 * time.Minute, Wait: 30 * time.Second}, t, func(r *retry.R) {
+					taskListOut := shell.RunCommandAndGetOutput(t, shell.Command{
+						Command: "aws",
+						Args: []string{
+							"ecs",
+							"list-tasks",
```
pglass, line 1045: non-blocking: It would be good to add some kind of helper function in `acceptance/framework/helpers` that runs the `ecs list-tasks` command and parses the output (I thought we had one, but no).

pglass (commented):
I have a bunch of small comments on the input variables, but this looks great!

modules/controller/variables.tf
pglass: Oh, I forget we are using `consulServers.hosts` in the ECS config json: https://github.com/hashicorp/consul-ecs/blob/ganeshrockz/agentless/config/schema.json#L85

Then maybe we should match Consul K8s and go with `consul_hosts` or `consul_server_hosts` (I like `hosts` because it indicates the port is not included)

modules/controller/variables.tf
```diff
@@ -0,0 +1,171 @@
...
+  default     = false
+}
+
+variable "tls_server_name" {
+  description = "The server name to use as the SNI host when connecting via TLS for Consul's HTTP and gRPC interfaces."
+  type        = string
+  default     = ""
+}
+
+variable "ca_cert_file" {
```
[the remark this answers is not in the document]
Ganeshrockz: Yeah that makes sense. I just exposed it because we allow configuring TLS CA cert via a file as well. I'll remove the same

modules/controller/variables.tf
Ganeshrockz: `consul_server_hosts` sounds lot more better

test/acceptance/tests/basic/basic_test.go
Ganeshrockz: Fixed it

modules/controller/variables.tf
```diff
@@ -0,0 +1,171 @@
...
+  description = "List of additional policy ARNs to attach to the execution role."
+  type        = list(string)
+  default     = []
+}
+
+variable "http_tls_config" {
+  type        = any
+  default     = {}
+  description = <<-EOT
+  This accepts HTTP specific TLS configuration based on the `consulServers.http` schema present in https://github.com/hashicorp/consul-ecs/blob/main/config/schema.json.
```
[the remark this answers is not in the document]
Ganeshrockz: Sure. I'll keep a note of this

Ganeshrockz:
@pglass I've addressed almost all the comments except the one that talks about the version. Please re-review it whenever possible. Thanks

test/acceptance/setup-terraform/main.tf
```diff
@@ -5,6 +5,10 @@ provider "aws" {
   region = var.region
 }
 
+provider "hcp" {
+  project_id = "6a513c39-907d-4a92-9159-3b03bd00ecda"
```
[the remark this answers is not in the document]
cthain: :thinking: I can't run the acceptance tests. I get an error during `terraform apply` that looks related to this.

```shell
setup-terraform$ terraform apply ...
...
╷
│ Error: unable to fetch project "6a513c39-907d-4a92-9159-3b03bd00ecda": [GET /resource-manager/2019-12-10/projects/{id}][403] ProjectService_Get default  &{Code:7 Details:[] Message:}
│ 
│   with provider["registry.terraform.io/hashicorp/hcp"],
│   on main.tf line 8, in provider "hcp":
│    8: provider "hcp" {
│ 

```

test/acceptance/tests/validation/validation_test.go
```diff
@@ -0,0 +1,906 @@
...
+
+	terraformOptions := &terraform.Options{
+		TerraformDir: "./terraform/volume-variable",
+		Vars:         map[string]interface{}{"volumes": volumes},
+		NoColor:      true,
+	}
+	t.Cleanup(func() {
+		_, _ = terraform.DestroyE(t, terraformOptions)
+	})
+	terraform.InitAndPlan(t, terraformOptions)
```
cthain, line 67: Should we be validating the output of the plan? This test only checks that the plan works (which is a good start) but it doesn't verify that the plan matches our expectations.

This is already a huge PR so maybe this can be addressed in a follow up?

cthain (approved):
Awesome job on this @Ganeshrockz! I left a few minor comments/suggestions. Nothing blocking.

test/acceptance/tests/validation/terraform/envoy-readiness-port-validate/main.tf
```diff
@@ -15,7 +14,7 @@ module "test_client" {
   container_definitions = [{
     name = "basic"
   }]
-  retry_join    = ["test"]
-  outbound_only = true
-  checks        = jsondecode(file("${path.module}/${var.checks_file}"))
-}
+  consul_server_hosts  = "consul.dc1"
```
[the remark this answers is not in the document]
cthain: Actually, there is a mix of usage throughout. This is not important to update so I'm resolving this comment.

test/acceptance/setup-terraform/main.tf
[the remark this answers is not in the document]
pglass: Projects seem to be a recent-ish addition to HCP. Or at least having multiple projects is.

The Terraform provider indicates that you may need to set project_id in some cases:

> Note:
If a [Project-Level Service Principal](https://cloud.hashicorp.com/docs/hcp/admin/iam/service-principals) is used, specify the default project_id in your provider configuration.

We probably need to add an input variable to setup-terraform so that we can specify an HCP project id if necessary. What do you think?

test/acceptance/tests/validation/validation_test.go
pglass: Most of the validation tests here were just moved to this new location (which is great 👌). So agreed, but I feel like improvements to the existing validation tests can wait for a follow up.

PR #193, opened 2023-08-07
Modify admin-partitions example to support dataplane

Ganeshrockz:
- Modified admin partitions example to support dataplane

## How I've tested this PR:

Manual deployment

## Checklist:
- [ ] Tests added N/A
- [ ] CHANGELOG entry added 


examples/admin-partitions/terraform/aws_ecs.tf
```diff
@@ -36,6 +36,22 @@ module "vpc" {
...
+      cidr_blocks = "0.0.0.0/0"
+    }
+  ]
+  default_security_group_ingress = [
+    {
+      from_port   = 0
+      to_port     = 0
+      protocol    = "-1"
+      cidr_blocks = "0.0.0.0/0"
+    }
```
pglass, line 53: The ingress rule should not allow traffic from any CIDR by default, to be safe. We can make this a self rule, so that only ingress from the same security group is allowed by default:

```
    {
      description = "Allow all"
      protocol = -1
      self = true
    }
```

But I think I'd prefer we remove these custom default security groups and try setting `manage_default_security_group = false` to see if that prevents Terraform from clobbering the AWS default rules. Up to you though.

examples/admin-partitions/terraform/variables.tf
```diff
@@ -15,13 +20,13 @@ variable "tags" {
 variable "consul_image" {
   description = "Consul Docker image."
   type        = string
-  default     = "public.ecr.aws/hashicorp/consul-enterprise:1.12.2-ent"
+  default     = "public.ecr.aws/hashicorp/consul-enterprise:1.16.0-ent"
```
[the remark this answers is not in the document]
Ganeshrockz: That makes sense. Removed consul_image and added a variable for dataplane image

pglass (approved):

examples/admin-partitions/terraform/aws_ecs.tf
Ganeshrockz: Making `manage_default_security_group` as `false` seems to be not altering the default security rules in the VPC. Going with that

PR #214, opened 2023-09-28
Reference ECS release candidate images for examples and submodules

Ganeshrockz:
- Bump ECS image variables to refer to the latest release candidate's name
- Split existing acceptance test matrix into Fargate and EC2 to reduce flakiness

## How I've tested this PR:
CI

👁️ 

## Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/examples/admin-partitions/terraform/variables.tf
+++ b/examples/admin-partitions/terraform/variables.tf
@@ -20,7 +20,7 @@ variable "tags" {
 variable "consul_ecs_image" {
   description = "Consul ECS image to use."
   type        = string
-  default     = "hashicorpdev/consul-ecs:latest"
+  default     = "hashicorppreview/consul-ecs:0.7.0-rc1"
 }
 
 variable "consul_dataplane_image" {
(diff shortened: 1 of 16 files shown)


examples/admin-partitions/terraform/variables.tf
```diff
@@ -20,7 +20,7 @@ variable "tags" {
 variable "consul_ecs_image" {
   description = "Consul ECS image to use."
   type        = string
-  default     = "hashicorpdev/consul-ecs:latest"
+  default     = "hashicorppreview/consul-ecs:0.7.0-rc1"
```
absolutelightning, line 23: can we define a variable 
```
variable "default_consul_ecs_image" {
  default = "hashicorppreview/consul-ecs:0.7.0-rc1"
}
```
and use it here like this - 
```
variable "consul_ecs_image" {
  type        = string
  default     = "hashicorpdev/consul-ecs:latest"
  default     = var.default_consul_ecs_image
}
```
or this is not suitable here? I see this repeated lot of times so I am suggesting.

Ganeshrockz: The method you proposed won't work across submodules. In the current state, we define one variable per submodule/terraform example.

absolutelightning: ok got it.

absolutelightning (approved):
LGTM!

PR #241, opened 2023-12-15
Fix health check command for mesh and gateway task submodules

Ganeshrockz:
- The ECS health check logic relies on the command returning a non zero exit code. In the current state, even when `consul-ecs-control-plane` returns back a `500` the `curl` command treats it a successful response and returns a non zero code. Here's an example of the same

```zsh
ganeshseetharaman@ganeshseetharaman-DFCQN60JDW % curl -v httpbin.org/status/500 
*   Trying 3.95.102.170:80...
* Connected to httpbin.org (3.95.102.170) port 80 (#0)
> GET /status/500 HTTP/1.1
> Host: httpbin.org
> User-Agent: curl/8.1.2
> Accept: */*
> 
< HTTP/1.1 500 INTERNAL SERVER ERROR
< Date: Fri, 15 Dec 2023 08:26:54 GMT
< Content-Type: text/html; charset=utf-8
< Content-Length: 0
< Connection: keep-alive
< Server: gunicorn/19.9.0
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Credentials: true
< 
* Connection #0 to host httpbin.org left intact
ganeshseetharaman@ganeshseetharaman-DFCQN60JDW % echo $?                                                          
0
```

This caused the `consul-dataplane` container to start up before the consul binary became available in the shared volume. In an ideal world, consul-dataplane should only start up after the control-plane container writes the Consul ECS binary to the shared volume and returns back a `200` response code for the `/consul-ecs/health` endpoint.

This PR fixes the same by adding a `-f` flag to curl.  

```zsh
ganeshseetharaman@ganeshseetharaman-DFCQN60JDW % curl -v -f httpbin.org/status/500
*   Trying 18.214.18.233:80...
* Connected to httpbin.org (18.214.18.233) port 80 (#0)
> GET /status/500 HTTP/1.1
> Host: httpbin.org
> User-Agent: curl/8.1.2
> Accept: */*
> 
< HTTP/1.1 500 INTERNAL SERVER ERROR
< Date: Fri, 15 Dec 2023 08:30:02 GMT
< Content-Type: text/html; charset=utf-8
< Content-Length: 0
< Connection: keep-alive
< Server: gunicorn/19.9.0
< Access-Control-Allow-Origin: *
< Access-Control-Allow-Credentials: true
* The requested URL returned error: 500
* Closing connection 0
curl: (22) The requested URL returned error: 500
ganeshseetharaman@ganeshseetharaman-DFCQN60JDW % echo $?                          
22
```

## How I've tested this PR:

Manual deployment

## Checklist:
- [ ] Tests added n/a
- [X] CHANGELOG entry added

### DIFF
--- a/modules/gateway-task/main.tf
+++ b/modules/gateway-task/main.tf
@@ -105,7 +105,7 @@ resource "aws_ecs_task_definition" "this" {
               initProcessEnabled = true
             }
             healthCheck = {
-              command  = ["CMD-SHELL", "curl localhost:10000/consul-ecs/health"] # consul-ecs-control-plane exposes a listener on 10000 to indicate it's readiness
+              command  = ["CMD-SHELL", "curl -f localhost:10000/consul-ecs/health || exit 1"] # consul-ecs-control-plane exposes a listener on 10000 to indicate it's readiness
               interval = 30
               retries  = 10
               timeout  = 5
--- a/modules/gateway-task/main.tf
+++ b/modules/gateway-task/main.tf
@@ -105,7 +105,7 @@ resource "aws_ecs_task_definition" "this" {
               initProcessEnabled = true
             }
             healthCheck = {
-              command  = ["CMD-SHELL", "curl -f localhost:10000/consul-ecs/health || exit 1"] # consul-ecs-control-plane exposes a listener on 10000 to indicate it's readiness
+              command  = ["CMD-SHELL", "curl -f localhost:10000/consul-ecs/health"] # consul-ecs-control-plane exposes a listener on 10000 to indicate it's readiness
               interval = 30
               retries  = 10
               timeout  = 5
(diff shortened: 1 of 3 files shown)


modules/gateway-task/main.tf
```diff
@@ -105,7 +105,7 @@ resource "aws_ecs_task_definition" "this" {
               initProcessEnabled = true
             }
             healthCheck = {
-              command  = ["CMD-SHELL", "curl localhost:10000/consul-ecs/health"] # consul-ecs-control-plane exposes a listener on 10000 to indicate it's readiness
+              command  = ["CMD-SHELL", "curl -f localhost:10000/consul-ecs/health || exit 1"] # consul-ecs-control-plane exposes a listener on 10000 to indicate it's readiness
```
v-rosa, line 108: the `|| 1` isn't redundant after adding the curl flag `-f`?

Ganeshrockz: Yes it is. I can remove the same.

lkysow (approved):

lkysow:
Happy to merge this quickly but we need to look into why the acceptance tests didn't catch this.

Ganeshrockz:
> Happy to merge this quickly but we need to look into why the acceptance tests didn't catch this.

I am pretty sure something changed recently in ECS. I have never seen a single acceptance test run fail with this error until last week. Right now I am pretty much able to reproduce this everytime I trigger the same suite of acceptance tests. I am also constantly seeing this error whenever trying to deploy the examples present in this repo.

PR #255, opened 2023-12-26
Changes to add health-sync container to the task definition

Ganeshrockz:
- Adds the `consul-ecs-health-sync` container to the mesh-task and gateway-task task definitions.
- Adds changes to the README to reflect the same.
- Needed to support https://github.com/hashicorp/consul-ecs/pull/211

## Checklist:
- [X] Tests added
- [X] CHANGELOG entry added 

### DIFF
--- a/modules/mesh-task/main.tf
+++ b/modules/mesh-task/main.tf
@@ -35,7 +35,7 @@ locals {
   namespace_tag = var.consul_namespace != "" ? { "consul.hashicorp.com/namespace" = var.consul_namespace } : {}
 
   // container_defs_with_depends_on is the app's container definitions with their dependsOn keys
-  // modified to add in dependencies on consul-ecs-control-plane and consul-dataplane.
+  // modified to add in dependencies on consul-ecs-mesh-init and consul-dataplane.
   // We add these dependencies in so that the app containers don't start until the proxy
   // is ready to serve traffic.
   container_defs_with_depends_on = [for def in var.container_definitions :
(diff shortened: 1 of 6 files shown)


modules/mesh-task/main.tf
```diff
@@ -46,10 +46,6 @@ locals {
           concat(
             lookup(def, "dependsOn", []),
             [
-              {
-                containerName = "consul-ecs-control-plane"
-                condition     = "HEALTHY"
-              },
               {
```
tauhid621, line 49: Should mesh-init be added here?

modules/mesh-task/variables.tf
```diff
@@ -145,7 +145,7 @@ variable "outbound_only" {
 variable "consul_ecs_image" {
   description = "consul-ecs Docker image."
   type        = string
-  default     = "hashicorppreview/consul-ecs:0.8.0-dev"
+  default     = "ganeshrockz/rearch-ecs"
```
tauhid621, line 148: Do we need this change? The other variable file has this as `default     = "hashicorppreview/consul-ecs:0.8.0-dev"`

tauhid621 (approved):
Looks good overall

modules/mesh-task/main.tf
Ganeshrockz: It's transitive. `dataplane` can only be healthy if `mesh-init` is successful and so this dependsOn block implicitly depends on `mesh-init`

modules/mesh-task/variables.tf
Ganeshrockz: Yes. Good catch

Ganeshrockz:
Ran the scenario validator CI for this PR branch and it turns out to be green https://github.com/hashicorp/terraform-aws-consul-ecs/actions/runs/7329432365

PR #259, opened 2024-01-02
[NET-6931]

kkavish:
- Enabled support to mount volumes for supporting TGWs
- Added the corresponding setting in variables.tf
- Added an example

## How I've tested this PR:
- Deployed using the new example created in `examples/terminating-gateway-tls`

## Checklist:
- [ ] Tests added
- [x] CHANGELOG entry added 


examples/terminating-gateway-tls/efs.tf
```diff
@@ -0,0 +1,152 @@
...
+  connection {
+    type        = "ssh"
+    user        = "ec2-user"
+    private_key = file(var.private_key)
+    host        = aws_instance.efs_mount_instance.public_ip
+    timeout     = "20s"
+  }
+
+
+  provisioner "remote-exec" {
```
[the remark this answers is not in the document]
Ganeshrockz: Through these examples we propose best practices to use Consul along with AWS components and this doesn't seem to reflect the same. One alternative approach I can think of is

1. Generate the cert PEMs via terraform and store the same in AWS secret manager
2. Add one additional container to the `external-server-app`'s task definition with the PEM values being passed as environment variables.
3. The container then starts up, writes the cert PEMs to the mounted EFS volume and exits successfully.
4. The server app's container starts up post that.

This way certs are handled in a safe and secure manner

examples/terminating-gateway-tls/efs.tf
```diff
@@ -0,0 +1,152 @@
+#resource "aws_key_pair" "deployer" {
```
Ganeshrockz, line 1: Are these commented out stuff in this file still relevant?

Ganeshrockz (changes requested):

examples/terminating-gateway-tls/external-server-app.tf
```diff
@@ -0,0 +1,440 @@
...
+    }
+  }
+
+  copy_cert_command = <<EOF
+cd /efs
+rm -rf *.cert
+rm -rf *.key
+echo "$TGW_EXTERNAL_APP_CA_CERT" > ./ca.cert
+echo "$TGW_EXTERNAL_APP_CERT" > ./gateway.cert
+echo "$TGW_EXTERNAL_APP_KEY" > ./gateway.key
```
Ganeshrockz, line 20: Why is the gateway's private key passed onto the server app?

examples/terminating-gateway-tls/efs.tf
```diff
@@ -0,0 +1,144 @@
...
+  name                    = "tgw-external-app-ca-key"
+  recovery_window_in_days = 0
+}
+
+resource "aws_secretsmanager_secret_version" "tgw_external_app_ca_key" {
+  secret_id     = aws_secretsmanager_secret.tgw_external_app_ca_key.id
+  secret_string = tls_private_key.tgw_external_app_ca_key.private_key_pem
+}
+
+resource "aws_secretsmanager_secret" "tgw_external_app_ca_cert" {
```
Ganeshrockz, line 77: Can you remove `tgw` prefix from the certs and keys related to the external apps? This is causing a lot of confusion when reading through the subsequent logic

examples/terminating-gateway-tls/efs.tf
```diff
@@ -0,0 +1,144 @@
...
+  recovery_window_in_days = 0
+}
+
+resource "aws_secretsmanager_secret_version" "tgw_external_app_ca_cert" {
+  secret_id     = aws_secretsmanager_secret.tgw_external_app_ca_cert.id
+  secret_string = tls_self_signed_cert.tgw_external_app_ca_cert.cert_pem
+}
+
+#generate cert and key for the gateway
+resource "tls_private_key" "tgw_external_app_private_key" {
```
Ganeshrockz, line 88: Similar to the previous comment, I would suggest to name this only `gateway_private_key` or something similar. The idea is to not club TGW and external app in a single name because they are two different entities.

examples/terminating-gateway-tls/README.md
```diff
@@ -0,0 +1,141 @@
...
+The plan should look similar to:
+
+```shell
+Plan: 96 to add, 0 to change, 0 to destroy.
+
+Changes to Outputs:
+  + consul_server_bootstrap_token = (known after apply)
+  + consul_server_lb_address      = (known after apply)
+  + mesh_client_lb_address        = (known after apply)
+```
```
Ganeshrockz, line 63: Some outputs in the `outputs.tf` file are missing here

examples/terminating-gateway-tls/README.md
```diff
@@ -0,0 +1,141 @@
+# Terminating Gateways on ECS.
```
Ganeshrockz, line 1:
```suggestion
# Terminating Gateways with mTLS on ECS.
```

examples/terminating-gateway-tls/README.md
```diff
@@ -0,0 +1,141 @@
...
+
+This example demonstrates accessing non mesh destinations services from mesh tasks with the help of terminating gateways deployed as ECS tasks.
```
Ganeshrockz, line 3: You could add information about how the terminating gateway performs mTLS with the external server app.

examples/terminating-gateway-tls/README.md
```diff
@@ -0,0 +1,141 @@
...
+      efs_volume_configuration = {
+          file_system_id = "<your_file_system_id>"
+          root_directory = "<path_inside_file_system_where_certs_are_present>"
+          iam = "ENABLED"
+        }
+    }
+]
+}
+```
+and set `tgw_certs_enabled = true` in `variables.tf` file.
```
Ganeshrockz, line 87: This part isn't needed. We need not allow users to configure these input variables.

_docs/terminating-gateway-external-server-tls.png
Ganeshrockz, line 1: I see that the connection between TGW and the external server app is marked as `consul mTLS`. This isn't exactly correct because these are custom mTLS connections between the app and TGW. Consul isn't the cert authority in this case.

Ganeshrockz: Typo: `Consu Server`

Ganeshrockz: It would be good to group the `consul-ecs-external-server-app` and `generate-and-import-cert-apps` inside a single ECS task. Right now they appear to be part of two different tasks in a single cluster.

Ganeshrockz: Should the arrow containing `Mount certs` point the other direction? Because we are mounting EFS onto the task and not the other way around. Please give a different to the arrow because it collides with that of the dataplane arrow's description.

Ganeshrockz: The legend still has references of Consul client. We don't use them anymore. Instead we use dataplane inside tasks. Can you make sure to update the diagram inside `consul-ecs-example-client-app` to reflect the same? You can refer to existing example READMEs for the same.

kkavish: > Should the arrow containing `Mount certs` point the other direction? Because we are mounting EFS onto the task and not the other way around. Please give a different to the arrow because it collides with that of the dataplane arrow's description.

Containers mounting volumes need to make an API call to EFS (pull), that's why the arrow is in the other direction. I will change the description to say "request to mount certs", would that help?

Ganeshrockz: That makes sense

examples/terminating-gateway-tls/external-server-app.tf
```diff
@@ -0,0 +1,440 @@
...
+resource "aws_security_group" "example_server_app_alb" {
+  name   = "${var.name}-external-server-app-alb"
+  vpc_id = module.vpc.vpc_id
+
+  ingress {
+    description = "Access to example server application."
+    from_port   = 0
+    to_port     = 0
+    protocol    = "-1"
+    cidr_blocks = ["0.0.0.0/0"]
```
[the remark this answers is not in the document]
kkavish: it's called from the fake service client, won't that call get blocked?
I'd restrict it to port 9090.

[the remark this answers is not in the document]
Ganeshrockz: Can we add the fake client's SG as the source security group to this?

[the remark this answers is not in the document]
Ganeshrockz: That and the `lb_ingress_ip` as well. We need to have 2 rules.

examples/terminating-gateway-tls/efs.tf
```diff
@@ -0,0 +1,47 @@
...
+resource "aws_security_group" "efs" {
+  name        = "${var.name}-efs-sg"
+  description = "Allows inbound efs traffic from ec2"
+  vpc_id      = module.vpc.vpc_id
+
+  ingress {
+    from_port   = 2049
+    to_port     = 2049
+    protocol    = "tcp"
+    cidr_blocks = ["0.0.0.0/0"]
```
Ganeshrockz, line 13: This should include the server app's SG as the source security group. Let's not expose stuff to the internet

examples/terminating-gateway-tls/README.md
Ganeshrockz: The outputs are still missing in this iteration

Ganeshrockz (commented):
Looks really close to approval. The stuff related to security group ingresses should be addressed for sure.

examples/terminating-gateway-tls/outputs.tf
```diff
@@ -0,0 +1,22 @@
...
+  value = module.dc1.dev_consul_server.bootstrap_token_id
+}
+
+output "mesh_client_lb_address" {
+  value = "http://${aws_lb.example_client_app.dns_name}:9090/ui"
+}
+
+output "certs_efs_file_system_address" {
+  value = aws_efs_file_system.certs_efs.id
+}
```
Ganeshrockz, line 18: Is this supposed to be a DNS name instead? I don't see any potential benefits in exposing the FS ID.

Ganeshrockz (approved):
Looks good to me. 🚀

PR #271, opened 2024-01-23
Support TProxy for gateway-task submodule

Ganeshrockz:
- Adds support for configuring TProxy for gateway-task submodule.

## How I've tested this PR:

CI

## Checklist:
- [X] Tests added
- [X] CHANGELOG entry added 


modules/gateway-task/main.tf
```diff
@@ -53,6 +53,61 @@ locals {
...
+          )
+        ] : [],
+        []
+      )
+    )
+  }
+
+  # Additional user attribute that needs to be added to run the mesh-init
+  # container with root access.
+  additional_user_attr                     = var.enable_transparent_proxy ? { user = "0" } : {}
```
absolutelightning, line 109: Question - When `var.enable_transparent_proxy` is false what will be the value of `user` in `finalized_mesh_init_container_definition` ?

Ganeshrockz: I think it will be empty or the user who starts the container. Definitely not `root`

PR #273, opened 2024-01-24
Add scenario validator

Ganeshrockz:
- Adds a workflow that runs scenario tests for a given scenario.

## Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 

### DIFF
--- a/.github/workflows/nightly-ecs-examples-validator.yml
+++ b/.github/workflows/nightly-ecs-examples-validator.yml
@@ -1,61 +1,21 @@
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
 name: Nighly ECS example validator
 on:
   workflow_dispatch:
 
 jobs:
   get-go-version:
-    runs-on: ubuntu-latest
-    defaults:
-      run:
-        working-directory: ./test/acceptance
-    outputs:
-      go-version: ${{ steps.get-go-version.outputs.go-version }}
-    steps:
-    - name: Checkout
-      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
-    - name: Determine Go version
-      id: get-go-version
-      run: |
-        echo "Building with Go $(cat .go-version)"
-        echo "go-version=$(cat .go-version)" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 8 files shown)


.github/workflows/scenario-validator.yml
```diff
@@ -0,0 +1,74 @@
...
+    outputs:
+      go-version: ${{ steps.get-go-version.outputs.go-version }}
+    steps:
+    - name: Checkout
+      uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+    - name: Determine Go version
+      id: get-go-version
+      run: |
+        echo "Building with Go $(cat .go-version)"
+        echo "go-version=$(cat .go-version)" >> "$GITHUB_OUTPUT"
```
absolutelightning, line 25: Should we extract this out in different file named `reusable-get-go-version.yaml` because this is also used in `.github/workflows/nightly-ecs-examples-validator.yml` ?

absolutelightning: I think we do this. Please check - https://github.com/hashicorp/consul-enterprise/blob/76b5de5039e9101416d5722046784e29d741d57f/.github/workflows/reusable-get-go-version.yml

Ganeshrockz: If its being done, then we should do it for other jobs as well like `terraform-fmt` and `go-fmt`. I have addressed this. Thanks for bringing this up.

.github/workflows/nightly-ecs-examples-validator.yml
```diff
@@ -1,61 +1,21 @@
...
-        args: "--verbose --enable gofmt"
-        only-new-issues: false
-        skip-pkg-cache: true
-        skip-build-cache: true
-        working-directory: ./test/acceptance
-    - name: Lint Consul retry
-      run: |
-        go install github.com/hashicorp/lint-consul-retry@v1.3.0
-        lint-consul-retry
+    - get-go-version
```
tauhid621, line 13: NIT: Shoud this have an indent?

tauhid621 (approved):

.github/workflows/nightly-ecs-examples-validator.yml
Ganeshrockz: No it does not since it's either way an array element

PR #282, opened 2024-02-07
Fix broken terminating-gtws with tls example

Ganeshrockz:
- Adds `enable_transparent_proxy` as `false` for fargate based tasks inside the term-gtw with tls example

## Checklist:
- [ ] Tests added
- [ ] CHANGELOG entry added 


kkavish (approved):
LGTM

Issue #318, opened 2024-05-22
v0.8.0 tf perpetual drift when enable_transparent_proxy is false


v-rosa:
Hey, we're testing the Consul ECS v0.8.0 terraform module, more specifically the mesh-task.

While testing we've found that after applying the changes, subsequent plans will keep displaying perpetual diffs in the container definitions about default values not being present. This doesn't happens in the v0.7.1 at least.

Terraform version: v1.3.7
AWS Provider version: v5.50.0
ECS deployment model: Fargate (meaning variable `enable_transparent_proxy` is set to false, this is relevant as per my findings).

Perpetual drift:

```
# module.ecs_services["consul-test-module-clt-test-task"].module.consul_mesh_template[0].aws_ecs_task_definition.this must be replaced
-/+ resource "aws_ecs_task_definition" "this" {
      ~ arn                      = "arn:aws:ecs:us-east-1:redacted:task-definition/stg-consul-test-module-clt-test-task-template:5" -> (known after apply)
      ~ arn_without_revision     = "arn:aws:ecs:us-east-1:redacted:task-definition/stg-consul-test-module-clt-test-task-template" -> (known after apply)
      ~ container_definitions    = jsonencode(
          ~ [ # forces replacement
              ~ {
                  - cpu              = 0 -> null
                    name             = "some-container-injected"
                  - portMappings     = [] -> null
                  - systemControls   = [] -> null
                  - volumesFrom      = [] -> null
                    # (6 unchanged elements hidden)
                } # forces replacement,
              ~ {
                    name             = "consul-dataplane"
                  - systemControls   = [] -> null
                    # (14 unchanged elements hidden)
                } # forces replacement,
              ~ {
                    name             = "consul-ecs-health-sync"
                  - systemControls   = [] -> null
                    # (13 unchanged elements hidden)
                } # forces replacement,
              ~ {
                  ~ linuxParameters  = {
                      ~ capabilities       = {
                          - add  = [] -> null
                          - drop = [] -> null
                        }
                        # (1 unchanged element hidden)
                    }
                    name             = "consul-ecs-mesh-init"
                  - portMappings     = [] -> null
                  - systemControls   = [] -> null
                    # (9 unchanged elements hidden)
                } # forces replacement,
              ~ {
                  - cpu              = 0 -> null
                    name             = "consul-test-module-clt-test-task"
                  - systemControls   = [] -> null
                  - volumesFrom      = [] -> null
[...]

PR #422, opened 2026-08-05
GH-318 Fixes a perpetual drift in terraform when enable_transparent_proxy is false on v0.10.0

v-rosa:

See https://github.com/hashicorp/terraform-aws-consul-ecs/issues/318

## How I've tested this PR:

As below.

Run tf apply and tf plan before the changes with enable_transparent_proxy  set to false, so you can verify the perpetual drift.

Run a new plan with the proposed changes and verify that the drift is gone.

### DIFF
--- a/modules/gateway-task/main.tf
+++ b/modules/gateway-task/main.tf
@@ -78,7 +78,10 @@ locals {
     ]
     linuxParameters = {
       initProcessEnabled = true
-      capabilities       = var.enable_transparent_proxy ? { add = ["NET_ADMIN"] } : {}
+      capabilities = {
+        add  = var.enable_transparent_proxy ? ["NET_ADMIN"] : []
+        drop = []
+      }
     }
     secrets = flatten(
       concat(
--- a/modules/mesh-task/main.tf
+++ b/modules/mesh-task/main.tf
@@ -96,7 +96,10 @@ locals {
     ]
     linuxParameters = {
       initProcessEnabled = true
-      capabilities       = var.enable_transparent_proxy ? { add = ["NET_ADMIN"] } : {}
+      capabilities = {
+        add  = var.enable_transparent_proxy ? ["NET_ADMIN"] : []
+        drop = []
+      }
     }
     secrets = flatten(
       concat(


v-rosa:
@krutibaraiya @paras-gupta2 @abhihashi can you please consider to be merged in the latest v0.10.x release?

LordAbhishek (commented):
Hi @v-rosa,
How about we remove the `capabilities` field from `linuxParameters` if transparent proxy is disabled, do you see any issue ?? I guess there should not be any drift. 

```
    linuxParameters = merge (
      { initProcessEnabled = true },
      var.enable_transparent_proxy ? { capabilities = { add = ["NET_ADMIN"] } } : {}
    )
```

v-rosa:
@LordAbhishek did you test it? if it works in the same way, we can go with your suggestion.

LordAbhishek:
Hi @v-rosa,
I tested it now. It works as expected. `capabilities` appear only when `transparent_proxy` is set to true.
Please do let me know if i need to test anything else or any other thought. Post that, I will update the PR.

**Testing:**
- First tab: No change and tf plan with `transparent_proxy` as false; `capabilities` still appears.
- Second tab: With above change and tf plan with `transparent_proxy` as false; `capabilities` do not appear at all.
- Third tab: With above change and tf plan with `transparent_proxy` as true; `capabilities` appear.

Screenshot 2026-09-09 at 9 46 59 AM

v-rosa:
So, I would add a final test, which is after the plan is applied, if you run a new plan, does it shows any drift?

LordAbhishek:
Hi @v-rosa,
I have tested re-plan as well. It doesn't drift. 
With transparent proxy disabled, it was drifting because of empty param's inside `capabilities`: `+ capabilities       = {}`. This field gets normalised to {add:[],drop:[]} when sent to ECS.

Now we have removed it, and it will be only present when transparent proxy is enabled as: 
```
                  + linuxParameters  = {
                      + capabilities       = {
                          + add = [
                              + "NET_ADMIN",
                            ]
                        }
                      + initProcessEnabled = true
                    }
```

PR #430, opened 2026-09-10
[DO NOT MERGE] compat check: consul-ecs main vs consul/consul-enterprise main

suresh-hashicorp:
## Purpose

**Do not merge.** Throwaway branch to get CI signal on whether **consul-ecs `main` works against Consul `main`** (CE and Enterprise). Opened as a draft so `terraform-ci` runs the acceptance suite.

## Why the usual approach doesn't cover this

Two separate gaps:

1. **consul-ecs is never tested from `main`.** `consul_ecs_image` is hardcoded to the released `public.ecr.aws/hashicorp/consul-ecs:0.10.0` in every module and test dir. Nothing overrides it — no test sets it and `TestConfig` has no field for it (unlike `consul_image`, which is built from `consul_version` and injected per-case).
2. **`consul_version` can't express `main`.** It's validated against `^\d+\.\d+\.\d+$`, and main reports `2.1.0-dev`.

So prior compat runs varied only the *released* Consul version against *released* consul-ecs `0.10.x`. This branch closes both gaps.

## Images under test

| Component | Version | Built from |
|---|---|---|
| consul | `2.1.0-dev` | `hashicorp/consul` main |
| consul-enterprise | `2.1.0-dev+ent` | `hashicorp/consul-enterprise` main |
| consul-ecs | `0.10.0-dev` | `hashicorp/consul-ecs` main |

Built `linux/amd64` and published to public Docker Hub repos. They must be public — the modules set no `repositoryCredentials`, so ECS cannot pull from a private registry. The Consul images layer the locally-built binary over the official base, mirroring `Consul-Dev.dockerfile` upstream.

## Changes

- `test/acceptance/framework/config/config.go` — `ConsulImageURI()` returns the dev images
- `modules/{mesh-task,controller,gateway-task}/variables.tf` — dev `consul_ecs_image` default; the `validation` tests inherit these
- `test/acceptance/tests/{basic,tproxy}/terraform/**` — same default, since these pass the variable through their own terraform dirs

HCP dirs and `examples/` are left alone (HCP acceptance is disabled as deprecated).

## Prior local result

The consul-ecs unit/integration suite — which execs a real `consul` binary from `PATH` — already passes fully against both CE and Enterprise `main`. That covers the Consul API surface (catalog, ACLs, namespaces/partitions, health sync). This PR adds the ECS runtime layer that can't be tested locally.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/modules/controller/variables.tf
+++ b/modules/controller/variables.tf
@@ -4,7 +4,8 @@
 variable "consul_ecs_image" {
   description = "consul-ecs Docker image."
   type        = string
-  default     = "public.ecr.aws/hashicorp/consul-ecs:0.10.0"
+  # COMPAT-CHECK (DO NOT MERGE): dev image built from consul-ecs `main`.
+  default = "docker.io/sureshkumardunga/consul-ecs-dev:main-amd64"
 }
 
 variable "ecs_cluster_arn" {
--- a/modules/dev-server/consul_1_14_compat.tf
+++ b/modules/dev-server/consul_1_14_compat.tf
@@ -4,5 +4,11 @@
 locals {
   // parse the consul version from the provided image: ["1", "12", "2"]
   consul_image_version_parts = regex(":.*(\\d+)[.](\\d+)[.](\\d+)", var.consul_image)
-  is_consul_1_14_plus        = tonumber(local.consul_image_version_parts[1]) >= 14
+  // Compare major and minor. Checking the minor alone misreads Consul 2.x as
+  // pre-1.14 (2.1.0 -> minor 1), which selects the legacy `ports.grpc` TLS
+  // config that Consul 2.x rejects at startup.
+  is_consul_1_14_plus = (
+    tonumber(local.consul_image_version_parts[0]) > 1 ||
+    (tonumber(local.consul_image_version_parts[0]) == 1 && tonumber(local.consul_image_version_parts[1]) >= 14)
+  )
 }
--- a/modules/gateway-task/variables.tf
+++ b/modules/gateway-task/variables.tf
@@ -108,7 +108,8 @@ variable "skip_server_watch" {
 variable "consul_ecs_image" {
   description = "consul-ecs Docker image."
   type        = string
-  default     = "public.ecr.aws/hashicorp/consul-ecs:0.10.0"
+  # COMPAT-CHECK (DO NOT MERGE): dev image built from consul-ecs `main`.
+  default = "docker.io/sureshkumardunga/consul-ecs-dev:main-amd64"
 }
 
 variable "consul_dataplane_image" {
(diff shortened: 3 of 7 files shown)


suresh-hashicorp:
## First run failed — root-caused to a real Consul 2.x bug in `modules/dev-server`

[Run 34439487633](https://github.com/hashicorp/terraform-aws-consul-ecs/actions/runs/34439487633) failed. Both `secure: true` cases of `TestBasic` (CE **and** Enterprise) timed out after 20 minutes waiting for the `consul-server` ECS service to reach steady state. `secure: false` passed.

The failure is in the Consul server container, before consul-ecs is ever exercised — so this is **not** a consul-ecs/Consul compatibility problem.

### Cause

`modules/dev-server/consul_1_14_compat.tf` compared only the **minor** version:

```hcl
is_consul_1_14_plus = tonumber(local.consul_image_version_parts[1]) >= 14
```

| Image | Parsed | `minor >= 14` |
|---|---|---|
| `consul:1.21.5` | `["1","21","5"]` | `21 >= 14` → true |
| `consul:2.1.0` | `["2","1","0"]` | `1 >= 14` → **false** |

Consul 2.x is misread as pre-1.14, so the server renders the legacy TLS config `ports { https = 8501, grpc = 8502 }` instead of `grpc_tls = 8503`. Consul 2.x refuses to start on that, reproduced locally:

```
==> the `ports.grpc` listener no longer supports TLS. Use `ports.grpc_tls` instead.
    This message is appearing because GRPC is configured to use TLS, but
    `ports.grpc_tls` is not defined
```

The container exits immediately, the service never stabilises, and the apply times out. With `grpc_tls = 8503` the same binary starts and elects a leader.

### Worth flagging beyond this PR

This is **not** an artifact of the dev image tag. Any Consul 2.x tag — including a future released `2.1.0` — hits it, so `dev-server` currently cannot deploy Consul 2.x with TLS. Probably deserves its own issue/PR against `main` independent of this compat check.

### Fix

280e0b8 compares major and minor. Verified across boundaries — `1.12.2`/`1.13.9` → false, `1.14.0` → true, `1.21.5`/`2.0.1`/`2.1.0` → true, so 1.x behaviour is unchanged.

suresh-hashicorp:
Compat check complete — consul-ecs `main` verified compatible with Consul and Consul Enterprise `main` on both Fargate and EC2 ([run 34446823174](https://github.com/hashicorp/terraform-aws-consul-ecs/actions/runs/34446823174)). Closing as intended; this branch was never meant to merge.

The `is_consul_1_14_plus` fix in 280e0b8 is a real bug affecting any Consul 2.x image with TLS and should be raised separately against `main`.

